PR #33, opened 2021-10-20
MTM-40337 - Add more precise the documentation for user delegation fu…

arkbsag:
MTM-40337 - Add more precise the documentation for user delegation functionality.


BeateRixen:
I did some language changes, @arkbsag please check if it is still correct.

PR #45, opened 2021-10-28
Remove CEP-Esper-related content

BeateRixen:


lukaszsoszynskisag (dismissed):

MWindel (dismissed):

kailol-sag (dismissed):

kailol-sag (approved):
Only looked at the update to apama/overview-analytics-bundle/streaming-analytics.md and this looks good to me.

BeateRixen:
> Only looked at the update to apama/overview-analytics-bundle/streaming-analytics.md and this looks good to me.
> 
> But the location of the CEL/Esper PDF in static/files/apama made me wonder, that is, the folder name "apama". This is not an Apama guide. Wouldn't it be less confusing if the PDF was located in a folder called something like "esper" or "deprecated"?

The folder is called Apama, since the link referencing the PDF is in the Apama folder. The same way as we do with images.

rssag (approved):

MWindel (approved):

PR #98, opened 2021-11-22
Feature/mtm 41682/alarm on non numeric measurements and no mapping known

Oezge-Akin:


sourajitsaha17 (approved):

MWindel (approved):

sag-eweingaertner:
@MWindel @BeateRixen Can you graft this also to 10.10 and 10.11? Or should we do that?

PR #134, opened 2021-12-15
New info about missing filter option in Device Management

MWindel:
Added information about the filtering option being unavailable in the main groups column but still available for filtering in the subgroup column.

Changes will be cherry-picked in 10.10., 10.11. and older releases as needed.

### DIFF
--- a/content/users-guide/device-management-bundle/grouping-devices.md
+++ b/content/users-guide/device-management-bundle/grouping-devices.md
@@ -27,7 +27,7 @@ Click **Groups** in the navigator to see all groups in a list format.
 
 For each group, various information is provided, for example the type and name. Click **Configure columns** at the right, to add or remove columns and customize the view to your preference. See also [Viewing devices > Configuring columns](/users-guide/device-management/#configuring-columns).
 
-To filter the groups for certain criteria, hover over the column headers and click the respective filter icon:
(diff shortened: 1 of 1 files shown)


janhommes (dismissed):

rssag (approved):

janhommes (approved):

content/users-guide/device-management-bundle/grouping-devices.md
```diff
@@ -27,7 +27,7 @@ Click **Groups** in the navigator to see all groups in a list format.
 
 For each group, various information is provided, for example the type and name. Click **Configure columns** at the right, to add or remove columns and customize the view to your preference. See also [Viewing devices > Configuring columns](/users-guide/device-management/#configuring-columns).
 
-To filter the groups for certain criteria, hover over the column headers and click the respective filter icon:
+To filter the groups for certain criteria, hover over the column headers and click the respective filter icon, see the screenshot below. However, filtering is not available for the main group column. The filtering option remains available in the subgroup column.
```
BeateRixen, line 30: remains > not sure if "remain" makes sense here? Perhaps better "The filtering option is only available in the subgroup column"?

MWindel: Yes you are right. I will change it to "is only available".

PR #140, opened 2021-12-17
Release notes for several PAQ versions for the Log4j issue

kailol-sag:


BeateRixen (approved):

rthrippleton (changes requested):
Everything is 100% fine, just waiting on the 10.6.6.x notes

rthrippleton (approved):

sharong143 (approved):
Confirmed that each version mentions the correct version of PAM, and the itrac are correct for each version, the text seems reasonable for the log4J issue and assuming that the styling when this would be displayed is correct

mjj29:
I approve of these, although I'm not 100% sure if this does the right thing

mjj29 (approved):
Approved

ddfourni1 (approved):

PR #204, opened 2022-01-19
image update

JoshuaAdemola:


BeateRixen (approved):

rssag (approved):

MWindel (approved):

BeateRixen:
@MWindel The latest screenshots from Joshua again show "Schwellwert". Strange enough, the names in the global smart rules overview page are not the same as those used in the specific smart rule as here the rule name is used. For example, we have "Alarm bei Erreichen eines expliziten Schwellwerts erzeugen" vs "Bei explizitem Schwellenwert Alarm erzeugen". Same in English. Seems to be attended.
@JoshuaAdemola For the sake of the screenshot, could you either remove the smart rule showing "Schwellwert" or correct in in the respective rule name field.

PR #257, opened 2022-02-14
bugherd task #535

rssag:


rssag:
the goal is to have the API landing page with the link still, but not have it as the front page, so users don't overlook the content that is actually still in the reference guide.

MWindel (approved):

PR #265, opened 2022-02-17
Edge. Removed month and year and replaced with version number.

sharath-jk:


BeateRixen:
@sharath-jk Note that this also needs to be done in older versions (back to 10.9 at least) and in the release notes.

MWindel (approved):

BeateRixen (approved):

PR #268, opened 2022-02-18
Updates images for German user guide

JoshuaAdemola:


BeateRixen:
> @JoshuaAdemola Looks good to me. I only found two issues: admin-application-properties: the page is cut off; I would show the entire page as in the previous screenshot. cockpit-smart-rules -list: Please show smart rules here which show a German name. Unfortunately, we are currently in a renaming process for the Ecosystem view, which means that about 10 screenshots might have to be retaken.

One more issue: The two screenshots admin-default-subscriptions... show English strings which have been replaced with the German equivalents in 10.13.44.0.

JoshuaAdemola:
> 



> @JoshuaAdemola Looks good to me. I only found two issues: admin-application-properties: the page is cut off; I would show the entire page as in the previous screenshot. cockpit-smart-rules -list: Please show smart rules here which show a German name. Unfortunately, we are currently in a renaming process for the Ecosystem view, which means that about 10 screenshots might have to be retaken.

Fixed

BeateRixen:
> > @JoshuaAdemola Looks good to me. I only found two issues: admin-application-properties: the page is cut off; I would show the entire page as in the previous screenshot. cockpit-smart-rules -list: Please show smart rules here which show a German name. Unfortunately, we are currently in a renaming process for the Ecosystem view, which means that about 10 screenshots might have to be retaken.
> 
> One more issue: The two screenshots admin-default-subscriptions... show English strings which have been replaced with the German equivalents in 10.13.44.0.

@JoshuaAdemola Could you please check if you can re-take the two images now (i.e. if they show German strings now)?

JoshuaAdemola:
Retook cockpit screenshots and updated admin-application-properties.

Still not able re-take admin-default-subscriptions

BeateRixen (approved):

BeateRixen:
@JoshuaAdemola I approved this PR but we should not yet merge it, but for test purposes first merge it into the German translations PR as soon as the translations are available.

MWindel (approved):

PR #281, opened 2022-02-23
Feature/mtm 42019/notifications2 doc review

rohitjoshi-softwareag:
#### https://cumulocity.atlassian.net/browse/MTM-42019

Apart from minor changes, I have removed the "Parallel consumers" section as it's not supported until at least 10.14. This section seeped into 10.11 documentation as well - so we need to explore what can be done about that.


sagscmi (approved):
Looks good to me.  Regarding the section on parallel consumers that was also accidentally included in the 10.11 doc, we should raise a defect & separate PR as soon as possible to resolve that.  Well spotted by the way!

BeateRixen (approved):

PR #304, opened 2022-03-09
Users guide update

JoshuaAdemola:


MWindel (approved):

content/users-guide/device-management-bundle/device-details.md
```diff
@@ -147,6 +147,12 @@ Details on these additional menu items are provided where required.
 
 The **Alarms** tab provides information on the alarms of a device. See [Working with alarms](#alarm-monitoring) for detailed information on alarms.
 
+<a name="availability"></a>
```
BeateRixen, line 150: This change should not go into 10.13. In 10.13 the tab is still called "Service monitoring", so the sorting should stay, also in the table at the top of this file.

JoshuaAdemola: Fixed. The section was removed.

PR #316, opened 2022-03-18
[MTM-43887] Grafting the changes to release 10.10

tsvet:
Grafting the changes from PR https://github.com/SoftwareAG/c8y-docs/pull/300 to 10.10.0

In the current implementation a user can only download 35MB of microservice log. 350MB of microservice log is wrong.
A fix will extend the 35MB to 115MB.
Remark: Downloading 115MB of microservice log will take 11 minutes, but documentation already states to use smaller portions.


hanneshofmann (approved):
LGTM, thanks!

I know that the PR description was copied from the initial PR but probably we could adjust the description (for all related PR's) because to my understanding there won't be a fix which is extending the limit.

MWindel (approved):

tsvet:
> LGTM, thanks!
> 
> I know that the PR description was copied from the initial PR but probably we could adjust the description (for all related PR's) because to my understanding there won't be a fix which is extending the limit.

Thanks Hannes. I adapted the description of the original PR and also here.

PR #326, opened 2022-03-21
Update Alarms topic to mention the streaminganalytics category

kailol-sag:
This is also to be cherry-picked into the 10.13 release.


rthrippleton (approved):

rthrippleton (changes requested):
"Such a change requires a restart of the Apama-ctrl microservice" (sorry, can't tag that line directly)
still implies it's manual, when it's actually fully automatic.

rthrippleton (approved):

PR #356, opened 2022-04-08
Image review and update. Removed plus button in certain screenshots

JoshuaAdemola:


MWindel (approved):

BeateRixen:
@JoshuaAdemola The image admin-user-totp-disabled is only shown in English. What again is the reason why you cannot take it in German?

BeateRixen:
@MWindel The image devmgmt-register-devices-custom shows a link to the "Devices guides". These do no longer exist. Can you please clarify with Pawel where to change this? Also I´m not sure where the link should point to.

JoshuaAdemola:
> @JoshuaAdemola The image admin-user-totp-disabled is only shown in English. What again is the reason why you cannot take it in German?

@BeateRixen The image can only be created in a staging environment that allows for TOTP. The staging environment I have only allows for SMS based two-factor authentication.

BeateRixen (approved):

PR #378, opened 2022-04-28
deprecation banner for 10.7.0

rssag:


BeateRixen:
@rssag I am actually not sure, if 10.7 is still in maintenance. We recently had a 10.7 security update. I will check.

BeateRixen (approved):

BeateRixen:
Checked with Jens, 10.7 is out of maintenance, so we can remove it from the dropdown and add the banner.

MWindel (approved):

PR #380, opened 2022-04-28
Edge/1010 revert changes

sharath-jk:
Replicating the changes from Edge 10.9 release.


rssag (approved):
i reintroduced the shortcodes, fixed quote/tick characters, fixed some markup. approved. please check if the changes are still what you need @sharath-jk

PR #388, opened 2022-05-04
[MTM-45303] document toggle in duplicate application modal

reey:
This PR adds additional info regarding the newly introduced switch in the duplicate application modal.


BeateRixen:
@reey Is this your 1st contribution to our docs? It looks really great :-). I´m only wondering why the image has a different size compared to the similar ones further above. May be, the size has changed in the UI?

BeateRixen (approved):

reey:
@BeateRixen yes, first contribution 😅 
I'm not really aware of how these images are usually taken, maybe I did something wrong there?
I guess the dialog received a bit more height because of the additional toggle that was added..
In addition to that it could be that I'm using another screensize than the person how previously created these..

BeateRixen:
> @BeateRixen yes, first contribution 😅 I'm not really aware of how these images are usually taken, maybe I did something wrong there? I guess the dialog received a bit more height because of the additional toggle that was added.. In addition to that it could be that I'm using another screensize than the person how previously created these..

Did you look into this document: https://docs.google.com/document/d/1geBLxk35ZIiS8C2Uo15H-14oneuEEKRf0eerHqdm3fw/edit#heading=h.bqw50lhopz2p?
The resolution looks good to me (72), I checked it. The browser size might be wrong but this should not have any effect since this dialog box should always show the original size. So I assume the original size changed, which means we cannot influence this here.

MWindel (approved):

janhommes (approved):

PR #389, opened 2022-05-05
Replaced faulty screenshot in Benutzerhandbuch

MWindel:
I replaced the faulty "Schwellenwert"-screenshot in the Smart rules collection section of the German User guide (Benutzerhandbuch).


kailol-sag (approved):

MWindel:
Screenshot cannot be replaced in 10.11 since new one is from 10.13

PR #421, opened 2022-05-18
Created DTM user-guide

JoshuaAdemola:


content/dtm.md
```diff
@@ -0,0 +1,8 @@
+---
+title: Cumulocity IoT DTM guide
```
BeateRixen, line 2: Cumulocity IoT DTM guide > Digital Twin Manager guide

content/dtm/localization.md
```diff
@@ -0,0 +1,7 @@
+---
+weight: 40
+layout: bundle
+title: Localization
+---
+
+Localization
```
BeateRixen, line 7: This can be removed, can´t it? It is already the title. Also applies to the other files

PR #454, opened 2022-06-14
bugherd task #576

rssag:


nneuerburg (approved):
The changes look good to me. I added feedback for adding a glossary entry for Device Protocols.
Additional Feedbcak: I am not a big fan of using the term "Custom Decoder Microservice", as we introduce a new concept here that we should also explain (what is a decoder). Instead, we could refer to a typcial microservice that implements a "Custom Device Protocol".

BeateRixen:
> The changes look good to me. I added feedback for adding a glossary entry for Device Protocols. Additional Feedbcak: I am not a big fan of using the term "Custom Decoder Microservice", as we introduce a new concept here that we should also explain (what is a decoder). Instead, we could refer to a typcial microservice that implements a "Custom Device Protocol".

True. The naming change has been done in this PR: https://github.com/SoftwareAG/c8y-docs/pull/422/files, i.e. "custom decoder microservice" or "custom codec microservice" has been changed to "custom device protocol". Probably this one slipped through.

rssag:
> The changes look good to me. I added feedback for adding a glossary entry for Device Protocols. Additional Feedbcak: I am not a big fan of using the term "Custom Decoder Microservice", as we introduce a new concept here that we should also explain (what is a decoder). Instead, we could refer to a typcial microservice that implements a "Custom Device Protocol".

@miryalarahul please have a look. the name of the microservice is "custom codec microservice", and is referred to as such many times throughout this file and the next three (that are linked to from the overview). i'm not sure there is an easy way of the renaming @nneuerburg sugggested ?

miryalarahul (approved):

PR #472, opened 2022-06-21
Replace English images with German images in the German User guide

JoshuaAdemola:


BeateRixen:
@JoshuaAdemola Regarding the image sso-user-data-mapping I´m not sure if this shows the correct content, looks quite different from the old one. Also, the title has changed and does no longer match the text where it should go to. Please compare to the English one.

BeateRixen:
@micst96 Could you please check the screenshots here as well? At least the last couple of images, showing the English UI, please.

BeateRixen:
> > Member
> 
> @BeateRixen, the image sso-user-data-mapping is correct and it matches the text in my Hugo preview. It looks different from the old one because the old one is wrong.

Ok, let´s merge it for now. I have created separate tickets to review the authentication screenshots anyway.

BeateRixen (approved):

PR #484, opened 2022-06-30
[MTM-46037] Introducing the overwrite parameter in the configuration …

tsvet:
This pull request refers to https://cumulocity.atlassian.net/browse/MTM-46037
A new manifest parameter has been introduced. It controls whether an option can be overwritten upon microservice update.

### DIFF
--- a/content/microservice-sdk/concept-bundle/manifest.md
+++ b/content/microservice-sdk/concept-bundle/manifest.md
@@ -229,6 +229,7 @@ The snapshot postfix means that the image build is a snapshot of your applicatio
 |:---|:---|:----------|:----------|
 |key|String |Key of the option|Yes
 |defaultValue|String|Default value|Yes
+|overwriteOnUpdate|Boolean|Defines if the option is reset upon microservice update <br>Default: true |No
 |editable|Boolean|Defines if the option can be changed by a subscribed tenant on runtime <br>Default: false |No
 
 #### Probe
--- a/content/microservice-sdk/concept-bundle/manifest.md
+++ b/content/microservice-sdk/concept-bundle/manifest.md
@@ -229,9 +229,8 @@ The snapshot postfix means that the image build is a snapshot of your applicatio
 |:---|:---|:----------|:----------|
 |key|String |Key of the option|Yes
 |defaultValue|String|Default value|Yes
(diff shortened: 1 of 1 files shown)


namnhatpham (approved):

content/microservice-sdk/concept-bundle/manifest.md
```diff
@@ -229,6 +229,7 @@ The snapshot postfix means that the image build is a snapshot of your applicatio
 |:---|:---|:----------|:----------|
 |key|String |Key of the option|Yes
 |defaultValue|String|Default value|Yes
+|overwriteOnUpdate|Boolean|Defines if the option is reset upon microservice update <br>Default: true |No
```
hanneshofmann, line 232: If I am not mistaken `overwriteOnUpdate` is only honored for editable settings and we probably should document this. We could move the new configuration below the `editable` documentation and could describe it as follows (or similar):

```suggestion
|overwriteOnUpdate|Boolean|Defines if an editable option is reset upon microservice update <br>Default: true |No
```

tsvet: You are right that this option only makes sense if we have an editable option. In terms of implementation, however, we do not distinguish whether the editable parameter is set or not. If we have a non-editable option, it does not matter actually, as it has not been changed (to be reset). Or am I missing something here?

hanneshofmann: > In terms of implementation, however, we do not distinguish whether the editable parameter is set or not.

Don't we check if `overwriteOnUpdate` **and** `editable` is set to check whether the option is saved/updated?

> If we have a non-editable option, it does not matter actually, as it has not been changed (to be reset). 

An option marked as editable means that a subscribed tenant at runtime can edit the option. Independently of this flag, a client in the owner scope of the microservice can always edit the option.

tsvet: Updated in 2ac2b9d860e721ded438a2ea041cee7183151e8e

hanneshofmann (approved):

BeateRixen (approved):

Markus-SWAG (approved):

BeateRixen:
@tsvet Can this be merged, and is this for 10.15 only?

PR #506, opened 2022-07-08
MTM-45721 update image when cloning applications.

eniosultan:


BeateRixen (approved):

BeateRixen:
@eniosultan Do we already use the new phrasing in the documentation (text)?

MWindel (approved):

eniosultan:
> @eniosultan Do we already use the new phrasing in the documentation (text)?

No, I didn't find any places where we use the old phrase, so only the image is updated.

PR #540, opened 2022-07-29
[DM-1096]: Added a missing explanation on why legacy configurations d…

dafkk:
…on't work with external URLs

### DIFF
--- a/content/reference/device-management-library-bundle/configuration.md
+++ b/content/reference/device-management-library-bundle/configuration.md
@@ -118,7 +118,9 @@ Devices that want to manage configuration as files can achieve a basic form usin
 
 This approach stores and transfers configuration as binary files.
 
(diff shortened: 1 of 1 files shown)


bozhidardedov (approved):

content/reference/device-management-library-bundle/configuration.md
```diff
@@ -118,6 +118,8 @@ Devices that want to manage configuration as files can achieve a basic form usin
 
 This approach stores and transfers configuration as binary files.
 
+> **Note:** This mechanism only works with internal Cumulocity repository. When a device supports only legacy configuration it is important to be known that configurations with external URLs will not be supported by that device. The id that is stored in the device managed object (an example can be seen below) is always refering to a inernal binary saved in the inventory.
```
BeateRixen, line 121:
```suggestion
{{< c8y-admon-info >}}
This mechanism only works with the internal {{< product-c8y-iot >}} repository. Be aware that configurations with external URLs will not be supported if a device only supports legacy configuration. The ID which is stored in the device managed object (an example can be seen below) always refers to an internal binary saved in the inventory.
{{< /c8y-admon-info >}}
```

BeateRixen (approved):

PR #544, opened 2022-08-03
Revert "DM-1109 Add documentation for image parameter of package plugin"

sag-eweingaertner:
Reverts SoftwareAG/c8y-docs#489

### DIFF
--- a/content/microservice-sdk/java-bundle/developing-microservice.md
+++ b/content/microservice-sdk/java-bundle/developing-microservice.md
@@ -336,7 +336,6 @@ It can be configured with the following parameters:
 
 * name (alias package.name) - defaults to project.artifactId
 * description (alias package.description) - defaults to project.description
-* image (optional, defaults to name) - name of the image to build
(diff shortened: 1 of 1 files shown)


sag-eweingaertner:
It turns out that we can't do this change, as it causes more problems than it solves.

BeateRixen (approved):

Oezge-Akin (approved):

content/microservice-sdk/java-bundle/developing-microservice.md
```diff
@@ -336,7 +336,6 @@ It can be configured with the following parameters:
 
 * name (alias package.name) - defaults to project.artifactId
 * description (alias package.description) - defaults to project.description
-* image (optional, defaults to name) - name of the image to build
```
manasesjesus, line 339: shouldn't the above xml/pom be updated as well?

```xml
            <configuration>
              <name>hello-world</name>
              <image>hello-world</image>              <!-- this one -->
              <encoding>UTF-8</encoding>
              <skip>false</skip>
            </configuration>
```

sag-eweingaertner: The image parameter was in the example before this change. It does not hurt and all our examples contain it in the pom. I'd leave it for now.

PR #573, opened 2022-08-23
Feature/DM-1382 reviewing viewing devices 10.15.0

dafkk:
Changes were missing from Viewing devices chapter:

- One was about still having the search in the data grid, which was an old functionality, that does not exists in the new grid, so it is removed.
- A small info popup on alarms was added for 10.15.0 to notify users that alarms shown on the list view are only current device alarms, not including the child alarms in this specific view. - I have changed the list image, not sure if it is correct, because it only contains my test devices. Let me know if we need to add a more accurate one or this one will work.


rssag (approved):

BeateRixen:
@dafkk @MWindel The changes in the German files will get overwritten by the translated files as soon as we add them but they do no harm here. Marlene, please take care that later on the removal of the entire section happens in the respective German file (we had some trouble with sections not being removed correctly in the past).

BeateRixen (approved):

dafkk:
Okay so I won't change geman text files. Just one more question do you need me to do the screenshot in german as well? @BeateRixen

PR #574, opened 2022-08-23
PAB-3323: Update links in C8Y doc to the new structure on the doc website

kailol-sag:
This is part of the changes for the automatic deployment of the DITA doc onto the doc website. The folder structure on the doc website will change for 10.15 when all docs for the October release will go live. Therefore I adapted all of the links in the Streaming Analytics guide and in the shortcodes. 

There was also one place in the Users guide and Benutzerhandbuch where I changed this.

Unfortunately, we cannot yet check if the changed links do really work. This has to wait until the middle of October 2022.

And this needs to be cherry-picked into the 10.15 branch.


BeateRixen (approved):

BeateRixen:
@kailol-sag Not sure if my approval is of much value here as I have no idea how the links must look like :-)

skom-sag (approved):

PR #579, opened 2022-08-24
[DM-1383]: Fixed outdated text and screenshots in grouping devices.

dafkk:


BeateRixen:
@dafkk In the user documentation, we generally use the terms "tab" and "page" and not view which is the technical term. We can discuss if we should change this but then we need to change this throughout the entire user documentation consistently. I prefer to differentiate between tabs and pages even if the technical implementation is the same (a view). Could you please discuss this with your team? We would then have to create an extra ticket for this change as this is quite some effort.

dafkk:
No, for the purpose of this change I wanted to change it not to be a tab anymore but a page. Just a view sounded better maybe. I can use page instead and there won't be any problems.

BeateRixen:
> No, for the purpose of this change I wanted to change it not to be a tab anymore but a page. Just a view sounded better maybe. I can use page instead and there won't be any problems.

OK, got it, then let´s go with page to be consistent for now. I will still discuss the "view" term with my team.

rssag (approved):

PR #584, opened 2022-08-25
[DM-1386]: Removed references to bulk operation tab in groups which i…

dafkk:
…s not existing anymore.


rssag (approved):

BeateRixen (approved):

BeateRixen:
@MWindel Again a heads up for you that this gets removed from the German files. We had the situation that removals of entire sections were not reflected when we got back the translations.

PR #592, opened 2022-08-30
DM-1439 Change Operations related screenshots for Actility

ashhSAG93:


BeateRixen:
The screenshot lora-devices-port has the wrong size. The browser size should be 1440px when taking screenshots. Moreover it shows your user name. Please show a neutral name like "admin" in the screenshots.

ashhSAG93:
> The screenshot lora-devices-port has the wrong size. The browser size should be 1440px when taking screenshots. Moreover it shows your user name. Please show a neutral name like "admin" in the screenshots.

Will make this change. Closing the PR as we do not want to merge directly into the release branch. Will create a PR out of develop and then we can backport the changes to 10.14 and 10.15 branches

PR #598, opened 2022-09-01
DM-1432 remove documentation about quick link widget

dominikschmitz:
As mentioned in the release notes of 10.14, we are deprecating and removing the quick link widget.
For that reason, we need to clean up the documentation.

@BeateRixen this needs to get grafted to 10.14 and 10.15, I guess


BeateRixen:
@dominikschmitz If we were precise we had to leave it in 10.14 and 10.15 with a deprecation note and only remove it from 10.16 onwards, as users can still see and use it in 10.14 and 10.15.

Oezge-Akin (approved):

BeateRixen (approved):

dominikschmitz:
> @dominikschmitz If we were precise we had to leave it in 10.14 and 10.15 with a deprecation note and only remove it from 10.16 onwards, as users can still see and use it in 10.14 and 10.15.

2 deprecation pr's for 10.14 and 10.15 are up for review.

PR #622, opened 2022-09-22
MTM-46186: Clarify messaging release support

sagscmi:
* MTM-46186 Clarify Notifications 2.0 availability
   - Available on SaaS environments for 10.13+
   - Available in dedicated/self-hosted environments for 10.11+
   - Not yet a 100% replacement for Real-time notifications
* MTM-46186 Clarify new Data Broker availability
   - Available on SaaS environments for 10.13+
   - Available in dedicated/self-hosted environments for 10.10+
   - Always needs to be enabled per-tenant


sagscmi:
This is a graft back from 10.16, tweaked slightly because we can as far as I know use `<admon>` tags in this release. I added only the same people who approved the original PR.  I wasn't sure if keeping the references to "10.13 and above" in the 10.11 documentation was sensible but but decided it was better than *not* saying where these feature are supported - given that customers (and internal people) are always asking this question.

slc-softwareag (approved):

BeateRixen (approved):

PR #675, opened 2022-11-15
removed note on Esper

BeateRixen:


kailol-sag (approved):

kailol-sag:
I did not find any other mentions of Esper in the user doc. And Richard T already took care of the link in https://github.com/SoftwareAG/apama-streaming-analytics-esper2apama . So I think we're good.

PR #692, opened 2022-11-29
Removed support user from feature list and fixed lists

BeateRixen:


content/users-guide/enterprise-tenant-bundle/overview.md
```diff
@@ -6,18 +6,17 @@ layout: redirect
 
 An {{< enterprise-tenant >}} offers additional administrative functionality compared to a {{< standard-tenant >}}, the major difference being **multi-tenancy**.
 
-Using an {{< enterprise-tenant >}}, you can
+Using an {{< enterprise-tenant >}}, you can:
```
rssag, line 9:
```suggestion
Using an {{< enterprise-tenant >}}, you can do the following:
```

BeateRixen: Same here, I would not be too strict. We do not always have to use "the following". Otherwise some sentences get somehow "clumsy". For me, a colon is also ok after "may be". We just agreed that it should not be in a middle of a sentence like "You can select if the customization of the home dashboard:"

rssag (approved):

MWindel (approved):

PR #696, opened 2022-12-01
New "parent_tenant_not_subscribed" alarm (PAB-3532)

kailol-sag:
I've added the new alarm as the last alarm in this topic. Is there a better location for this - after one of the other alarms?

Not sure about the terminology:
The alarm type includes "parent".
The alarm text, however, mentions the "enterprise tenant". 
Why are 2 different terms used here? What's the correct one?

Also: What's the correct terminology? "child tenant" or (as seen in the C8Y doc) "subtenant"? Hoping for feedback from the C8Y doc team here.

### DIFF
--- a/content/apama/troubleshooting-bundle/alarms.md
+++ b/content/apama/troubleshooting-bundle/alarms.md
@@ -399,7 +399,7 @@ To diagnose the cause, you can try the following. It may be that the Apama-ctrl 
 This alarm is raised for a subtenant that was subscribed before the parent tenant was subscribed.
 
 - Alarm type: `parent_tenant_not_subscribed`
-- Alarm text: Microservice functionality will not work until the enterprise tenant is also subscribed. Contact the administrator.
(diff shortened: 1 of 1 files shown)


BeateRixen:
> I've added the new alarm as the last alarm in this topic. Is there a better location for this - after one of the other alarms?
> 
> Not sure about the terminology: The alarm type includes "parent". The alarm text, however, mentions the "enterprise tenant". Why are 2 different terms used here? What's the correct one?
> 
> Also: What's the correct terminology? "child tenant" or (as seen in the C8Y doc) "subtenant"? Hoping for feedback from the C8Y doc team here.

The term "enterprise tenant" in the message is confusing to me. I suppose "parent tenant" would be better here. Regarding "subtenant vs child tenant", it is right, we use the term subtenant, but this has also already been controversially discussed ( to simply use "tenant" instead).

BeateRixen (approved):

MWindel (approved):

PR #710, opened 2022-12-05
Mtm 48945 doc rename applications to web apps

eniosultan:


BeateRixen:
@eniosultan @janhommes I´m not sure about this background of this change. If we do no longer speak about applications this would need to be reflected the entire documentation and not just the administration section of the User guide. Is this a general terminology change?

janhommes:
@BeateRixen this is coming from Niko. Ask him... I don't know.

BeateRixen:
@janhommes Will do. A simple replacement of each and all occurrences of the term "application" in only a small part of the documentation is probably not what he intended :-)

janhommes:
> @janhommes Will do. A simple replacement of each and all occurrences of the term "application" in only a small part of the documentation is probably not what he intended :-)

As far as I know only the Ecosystem -> Applications should be renamed to "Web apps":
image


@eniosultan you tried only to update this in the documentation, right?

BeateRixen:
Still we would need to change it in other places in the documentation as well, for example in the Concepts guide: https://cumulocity.com/guides/concepts/applications/. And actually in many other places. This PR simply replaces each occurrence of "application" with "web app". I suppose that we still speak about applications in general?

eniosultan:
This was the reason I created a draft PR, it was work in progress until we figure out exactly what do we want to change...
I'm really not sure to what extend do we want to change this. For example, are we going to call hosted applications, "hosted web applications" from now on? I need some more clarity so I guess we should ask Niko...

PR #722, opened 2022-12-12
Modified documentation part for file upload type in property screen.

Vagdevi1998:
Modified the description for file upload under property library screen.


lra-sag (approved):

MWindel:
Why is this only for 10.16.0?

PR #730, opened 2022-12-13
MTM-49961/MTM-47816 remove esper reference from microservice SDK…

jkot-softwareag:
…real-time notification example

### DIFF
--- a/content/microservice-sdk/rest-bundle/microservice-development.md
+++ b/content/microservice-sdk/rest-bundle/microservice-development.md
@@ -526,18 +526,12 @@ A status of PENDING means here that the device has not yet picked up the operati
 
 #### Listening for events
 
-Besides querying the {{< product-c8y-iot >}} data store, you can also process and receive events in real time as described in [Real-time processing in {{< product-c8y-iot >}}](/concepts/realtime). For example, assume that you would like to display real-time location updates on a map. Use the Administration application (or the [REST API](https://{{< domain-c8y >}}/guides/10.9.0/event-language/real-time-statements/)) to create a new rule module "myRule":
-
-```sql
-select *
-from EventCreated e
-where e.event.type = "c8y_LocationUpdate";
-```
+Besides querying the {{< product-c8y-iot >}} data store, you can also process and receive events in real time. For example, assume that you would like to display real-time location updates on a map.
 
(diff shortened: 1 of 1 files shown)


skom-sag (approved):

content/microservice-sdk/rest-bundle/microservice-development.md
```diff
@@ -560,16 +554,17 @@ HTTP/1.1 200 OK
 }]
 ```
 
-After the handshake, the client needs to subscribe to the output of the above rule. This is done using a POST request with the module name and the statement name as subscription channel. In our example, we used the module name "myRule" and did not give a name to the `select` statement ("@Name('')"), so the subscription channel is "/myRule/\*".
+After the handshake, the client needs to subscribe to the particular device channel or all devices channel (*). This is done using a POST request with the name of the channel.
```
rssag, line 557:
```suggestion
After the handshake, the client must subscribe to the particular device channel or all devices channel (*). This is done using a POST request with the name of the channel. 
```

content/microservice-sdk/rest-bundle/microservice-development.md
```diff
@@ -560,16 +554,17 @@ HTTP/1.1 200 OK
 }]
 ```
 
-After the handshake, the client needs to subscribe to the output of the above rule. This is done using a POST request with the module name and the statement name as subscription channel. In our example, we used the module name "myRule" and did not give a name to the `select` statement ("@Name('')"), so the subscription channel is "/myRule/\*".
+After the handshake, the client needs to subscribe to the particular device channel or all devices channel (*). This is done using a POST request with the name of the channel. 
+Let's assume that in our example we want to receive notifications about events from the device with internal id 24800. In this case the subscription channel is "/events/24800".
```
rssag, line 558:
```suggestion
For example, if you want to receive notifications about events from the device with internal ID "24800", the subscription channel is "/events/24800".
```

rssag (changes requested):

rssag (approved):

content/microservice-sdk/rest-bundle/microservice-development.md
jkot-softwareag: done

content/microservice-sdk/rest-bundle/microservice-development.md
jkot-softwareag: done

PR #755, opened 2022-12-30
BugherdTask_#612

dianaolives:
Changed http to https

Includes bugherd tasks 611 and 636

### DIFF
--- a/content/microservice-sdk/concept-bundle/manifest.md
+++ b/content/microservice-sdk/concept-bundle/manifest.md
@@ -16,7 +16,7 @@ Here is an example manifest:
     "version": "1.0.0",
     "provider": {
         "name": "New Company Ltd.",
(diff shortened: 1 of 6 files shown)


BeateRixen:
@rssag What do you think? Can we change all these occurrences from "http" into "https"? Or de we need to get back to some developer? Not sure though who this could be :-). This was a request via BH for some occurrences.

rssag:
> @rssag What do you think? Can we change all these occurrences from "http" into "https"? Or de we need to get back to some developer? Not sure though who this could be :-). This was a request via BH for some occurrences.

should be ok. one is an example link, api URLs should be anyway, domains around cumulocity too.

rssag (approved):

BeateRixen (approved):

PR #797, opened 2023-01-19
MTM-49613 Replaced screenshots with new layout for the german users g…

eniosultan:
…uide for cockpit


BeateRixen:
@eniosultan The images cockpit-app-configuration and cockpit-data-point-library show an English header. 
And the image cockpit-reports-navigator has a different focus. Intended?

eniosultan:
> @eniosultan The images cockpit-app-configuration and cockpit-data-point-library show an English header. And the image cockpit-reports-navigator has a different focus. Intended?

The focus does not really matter IMO as the location menu can be toggled from the  reports page. So the screenshot is still relevant

BeateRixen (approved):

janhommes (approved):

PR #837, opened 2023-02-03
BugherdTask_#673

dianaolives:
Fixed link


BeateRixen (approved):

BeateRixen:
@dianaolives Could you please check in which versions this change is relevant? I would then fix it back to 10.13.

PR #860, opened 2023-02-08
MTM-51330 trusted cert screenshots

mmed-softwareag:


Michal-Bak (approved):

BeateRixen:
@mmed-softwareag I would definitely delete the image "devmgmt-mqtt-cert-pop-completed" and change the text slightly. In general, some adaptions to the text are required if you change the scope of the images (which is good), so that the user know where he is. For example, "in the **Proof of possession** section in the the certificate details "... If you like, I can do that for you.

BeateRixen (approved):

PR #871, opened 2023-02-13
DM-1013 Fix properties in OPC UA failure handling

sag-eweingaertner:
DM-1013  Add auto address space setting to example config


sag-eweingaertner:
Wrong branch

PR #883, opened 2023-02-14
MTM-50748 Change screenshots for Microservice SDK for Java

karolkaczmarski-sag:
Not sure, but maybe ms-tracking-map.png can be removed? https://cumulocity.com/guides/10.16.0/microservice-sdk/java/#testing-the-application


Airdroplet:
I think that the screenshot to tracking application might be helpful to quickly visually find a feature UI offers so customers could easily test for correctness of events created by microservice. Alternative could be a text describing how to find it, but it would be less representative in my opinion. Probably better to maintain the screenshot in this case, since one look and I don't need to read the part describing the screenshot to understand what comes next.

Maybe it also worth to crop out `powered by Cumulocity` from screenshots. At least this idea was given by @pawel-rynarzewski-sag in one of other PRs.

Airdroplet (approved):

rssag (approved):

BeateRixen (approved):

jkot-softwareag (approved):

PR #931, opened 2023-02-27
[MTM-51784] Fixed missing sections in `index.json`s

pawel-rynarzewski-sag:
Incorrect list markup caused some sections to be missing from `index.json` files.

Before:

After:

Should be grafted to earlier releases, too.

### DIFF
--- a/content/users-guide/administration-bundle/managing-applications.md
+++ b/content/users-guide/administration-bundle/managing-applications.md
@@ -26,10 +26,10 @@ helpcontent:
 
 
     Click on an application to view the application details. To add an application, click **Add application** and follow the instructions in the wizard, see also the *User guide*."
-    label: packages
+  - label: packages
     title: Packages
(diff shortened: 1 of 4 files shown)


BeateRixen (approved):

BeateRixen:
To which versions should this be grafted? Probably 10.17 and 10.16? I think for 10.15 we did not have these entries but I will check.

pawel-rynarzewski-sag:
@BeateRixen, ideally to all where these entries were present. At least up until 10.16 (for that version we got a blocker).

catarino-sag (approved):

Oezge-Akin (approved):

BeateRixen:
> @BeateRixen, ideally to all where these entries were present. At least up until 10.16 (for that version we got a blocker).

Cherry-picked to 10.17 and 10.16. Not relevant in earlier versions.

PR #951, opened 2023-03-22
Revert "DM-2055: Removed the cluster tenant cleanup operation

sourajitsaha17:
Creating a new revert PR for this one because we decided to address this part with a separate ticket. From the backend perspective, we are still planning some improvements on that.


sourajitsaha17:
Sorry for the confusion. This part is fine.

PR #965, opened 2023-04-05
BugherdTask#668_10.17

dianaolives:
According to Philipp, the screenshots that were updated was missing functionalities from the latest version and needed to be updated. Also, like in release 10.14, I copied the german text from 10.16 so that it went along with the english version.


BeateRixen:
@dianaolives The German md file will be replaced by the one we get back from the translators. @MWindel Please double-check after we got back the translations for 10.17 that these changes are included.

rssag (approved):

PR #984, opened 2023-04-17
MTM-50341 add information which apps need to be subscribe for tenant …

jkot-softwareag:
…to become Enterprise Edition.

Preview of the change (I wanted to be precise with types of applications, since those are now on different UI screens)

image

### DIFF
--- a/content/users-guide/enterprise-tenant-bundle/overview.md
+++ b/content/users-guide/enterprise-tenant-bundle/overview.md
@@ -64,3 +64,11 @@ The following sections describe the following additional functionalities availab
 </tr>
 </tbody>
 </table>
+
+{{< c8y-admon-req >}}
+In order to make particular tenant to be **Enterprise tenant** you need to subscribe following 4 "applications" to the tenant (3 features and 1 microservice):
+* feature-branding
+* feature-broker
+* feature-user-hierarchy
+* sslmanagement
+{{< /c8y-admon-req >}}


Airdroplet (approved):

darluk (approved):

content/users-guide/enterprise-tenant-bundle/overview.md
```diff
@@ -64,3 +64,11 @@ The following sections describe the following additional functionalities availab
 </tr>
 </tbody>
 </table>
+
+{{< c8y-admon-req >}}
+In order to make particular tenant to be **Enterprise tenant** you need to subscribe following 4 "applications" to the tenant (3 features and 1 microservice):
```
karolkaczmarski-sag, line 69:
```suggestion
In order to make particular tenant to be **Enterprise tenant** you need to subscribe following applications to the tenant:
```

karolkaczmarski-sag (approved):

content/users-guide/enterprise-tenant-bundle/overview.md
jkot-softwareag: up to Beate and UI to decide, or better - for some consultants who work with customers. 

As I written in initial comment - I wanted to be precise, because "features" are on the different screen than "microservices". For BE those are all "applications", but in UI none of them is on web-applications screen (which I think is default when we think about **applications**)

BeateRixen:
@jkot-softwareag @Airdroplet I´m not sure if we want to share the information in this way with customers at all. I remember that we had several discussions on the difference between standard and enterprise tenants from a technical versus a commercial perspective. I will check with Niko and Jane.

jkot-softwareag:
> @jkot-softwareag @Airdroplet I´m not sure if we want to share the information in this way with customers at all. I remember that we had several discussions on the difference between standard and enterprise tenants from a technical versus a commercial perspective. I will check with Niko and Jane.

This was raised by Tobias Sommer from customer perspective (NTT I think, please see source of the ticket) https://cumulocity.atlassian.net/browse/MTM-50341
The problem is not with our 4 cloud environments `cumulocity.com`, but with those 50+ on-premise environments, operated by our Ops (cloud or MaSe) or by customer Ops. Multiple customers come with this question over and over again. Or raise support tickets that _"Enterprise Edition is not working, please help"_.

nneuerburg:
I believe that the procedure of upgrading a standard tenant to an enterprise tenant should rather be documented in the Ops guide (if this is not the case already) given that this cannot be done by normal users but only instance administrators. Also the description seems to be incomplete. Afaik, there needs to be a flag set in addition to allow to create sub-tenants.

jkot-softwareag:
> Also the description seems to be incomplete. Afaik, there needs to be a flag set in addition to allow to create sub-tenants.
The flag is automatically updated when last of 4 application is subscribed. 

No problem, I can close this PR as wont do, but can somebody at least contact customer success people and customer consultants? The ticket was raised by them based on history with customers.

BeateRixen:
> > @jkot-softwareag @Airdroplet I´m not sure if we want to share the information in this way with customers at all. I remember that we had several discussions on the difference between standard and enterprise tenants from a technical versus a commercial perspective. I will check with Niko and Jane.
> 
> This was raised by Tobias Sommer from customer perspective (NTT I think, please see source of the ticket) https://cumulocity.atlassian.net/browse/MTM-50341 The problem is not with our 4 cloud environments `cumulocity.com`, but with those 50+ on-premise environments, operated by our Ops (cloud or MaSe) or by customer Ops. Multiple customers come with this question over and over again. Or raise support tickets that _"Enterprise Edition is not working, please help"_.

The Ops guides are provided to the on-prem customers/operators as well. So it seems to be a good suggestion to add the information there instead of the user documentation.

PR #990, opened 2023-04-21
IOT-17056 link fix

rssag:


BeateRixen (approved):

rssag:
what are the target versions? @janhommes

PR #994, opened 2023-04-24
MTM-51494-Document-the-maximum-limitation-of-devices-per-group

Airdroplet:
* added recommendation to concept guide regarding asset hierarchy consideration in terms of JSON document size limitation

### DIFF
--- a/content/concepts/domain-model-bundle/inventory.md
+++ b/content/concepts/domain-model-bundle/inventory.md
@@ -87,8 +87,9 @@ For more information on fragments and how managed objects are structured, see th
 {{< c8y-admon-info >}}
 While designing the data model for the inventory managed object consider the following:
 1. There is no size or length constraint for a single fragment, but there is a limitation for the overall JSON document size, which may not exceed 16MiB for a single managed object entry within the inventory collection. We recommend you to keep it below 1 MiB.
-2. When you include arrays of elements within fragments, keep the length of such collections below 1k elements.
-3. Each consecutive fragment added to the managed object at root level imposes a certain delay on querying such an item. If the performance of a query matters, it is recommended to nest custom fragments with information within a chosen single fragment effectively limiting the root fragments number. For example:
+2. When designing assets hierarchy, be advised to use smaller groups. Each subitem in assets hierarchy creates a reference record in parent item. Therefor keep in mind 1st recommendation regarding JSON document size.
+3. When you include arrays of elements within fragments, keep the length of such collections below 1k elements.
+4. Each consecutive fragment added to the managed object at root level imposes a certain delay on querying such an item. If the performance of a query matters, it is recommended to nest custom fragments with information within a chosen single fragment effectively limiting the root fragments number. For example:
 ```json
 {
     "id": "47035",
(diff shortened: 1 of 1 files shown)


content/concepts/domain-model-bundle/inventory.md
```diff
@@ -87,8 +87,9 @@ For more information on fragments and how managed objects are structured, see th
 {{< c8y-admon-info >}}
 While designing the data model for the inventory managed object consider the following:
 1. There is no size or length constraint for a single fragment, but there is a limitation for the overall JSON document size, which may not exceed 16MiB for a single managed object entry within the inventory collection. We recommend you to keep it below 1 MiB.
-2. When you include arrays of elements within fragments, keep the length of such collections below 1k elements.
-3. Each consecutive fragment added to the managed object at root level imposes a certain delay on querying such an item. If the performance of a query matters, it is recommended to nest custom fragments with information within a chosen single fragment effectively limiting the root fragments number. For example:
+2. When designing assets hierarchy, be advised to use smaller groups. Each subitem in assets hierarchy creates a reference record in parent item. Therefor keep in mind 1st recommendation regarding JSON document size.
```
Airdroplet, line 90: Not sure if it is possible to precise it with some number. Each case is different and would depend on how customers are designing theirs solutions. Main idea is to tell that if customer see a request response payload from inventory API of size bigger than 1 MiB, it might be a good time to optimize hierarchy .

jkot-softwareag: I would leave somewhere this recommendation with 1K elements. Because there are people out there who add 100K elements to group and open tickets for us that "adding next item is slow", instead of creating e.g. 2 level hierarchy

Airdroplet: I agree, original problem was that group contained 200 K devices, and limit for MO exceeded.

content/concepts/domain-model-bundle/inventory.md
```diff
@@ -87,8 +87,9 @@ For more information on fragments and how managed objects are structured, see th
 {{< c8y-admon-info >}}
 While designing the data model for the inventory managed object consider the following:
 1. There is no size or length constraint for a single fragment, but there is a limitation for the overall JSON document size, which may not exceed 16MiB for a single managed object entry within the inventory collection. We recommend you to keep it below 1 MiB.
-2. When you include arrays of elements within fragments, keep the length of such collections below 1k elements.
-3. Each consecutive fragment added to the managed object at root level imposes a certain delay on querying such an item. If the performance of a query matters, it is recommended to nest custom fragments with information within a chosen single fragment effectively limiting the root fragments number. For example:
+2. When designing assets hierarchy, be advised to use smaller groups limiting it to 1000 subassets. Each subitem in assets hierarchy creates a reference record in parent item. Therefor keep in mind 1st recommendation regarding JSON document size.
```
BeateRixen, line 90:
```suggestion
2. When designing asset hierarchies, use small groups with less than 1000 subassets. Each subitem in the asset hierarchy creates a reference record in the parent item. Therefore keep in mind the first recommendation regarding the JSON document size.
```

karolkaczmarski-sag (approved):

jkot-softwareag (approved):

PR #999, opened 2023-04-25
Update configuring-karaf.md

sharath-jk:


sharath-jk:
Merging without second approval as the change is minor and was approved in another PR https://github.com/SoftwareAG/c8y-docs/pull/989

PR #1040, opened 2023-05-11
MTM-50640/MTM-53153-documentation-improvements

Airdroplet:
* Fixed typos
* Updated properties names

Details: 
MTM-50640/MTM-53153-documentation-improvements

[x]	https://cumulocity.com/guides/10.16.0/users-guide/administration/#tfa-settings SMS-based settings naming should be changed from.. to..

[-]	in https://cumulocity.com/guides/10.16.0/users-guide/administration/#managing-the-connectivity-settings are specified only: Actility LoRa, Sigfox, SIM. why to Impact and Loriot as well?
> Should check for input from Data in Control

[x]	https://cumulocity.com/guides/10.16.0/users-guide/administration/#how-to-enable-a-specific-user name of the paragraph is How to enable a specific user - would be better to change to How to enable for a specific user

[x]	in https://cumulocity.com/guides/10.16.0/users-guide/administration/#how-to-enable-a-specific-user for 2 point: `Select the checkbox next to Enable two-factor authentication.` better to rename to Two-factor authentication as is on design

[x]	in https://cumulocity.com/guides/10.16.0/users-guide/administration/#custom-template in New roles are added to the user from every matching access mapping. If one access mapping statement assigns the role “admin” and a second one assigns the role “business” and both meet the defined conditions, then the user will be granted access to the global roles “business” and “admin”." remove quotes at the end of the sentence

[x]	in https://cumulocity.com/guides/10.16.0/users-guide/administration/#custom-template Each access token is signed by a signing certificate. Currently there are three options to configure the signing certificates. on the page are described 4 options for signing certificates not 3

[-]	In part https://cumulocity.com/guides/10.16.0/users-guide/administration/#passwords in Password validity limit (days) - the number of days a password may be valid before it must be reset; minimum value is “0”, maximum value is “999999”. Leave empty to use the value from the tenant options configured in the Management tenant, see Cumulocity IoT Core - Operations guide. there is no link for Cumulocity IoT Core - Operations guide.
> it is so because operations guide is usually a pdf document available in knowledge center, for example: https://documentation.softwareag.com/cumulocity_iot_platform/iot10-15-0/index.htm

### DIFF
--- a/content/users-guide/administration-bundle/configuring-single-sign-on.md
+++ b/content/users-guide/administration-bundle/configuring-single-sign-on.md
@@ -146,7 +146,7 @@ On user login, user data like first name, last name, email and phone number can 
 
 Mapping for alias is not available because it is not used in the context of SSO login.
 
-Each access token is signed by a signing certificate. There are following options to configure the signing certificates.
+Each access token is signed by a signing certificate. The following options are available to configure the signing certificates.
 
 1. By specifying the Azure AD certificate discovery address.
 
--- a/content/users-guide/administration-bundle/configuring-single-sign-on.md
+++ b/content/users-guide/administration-bundle/configuring-single-sign-on.md
@@ -93,7 +93,7 @@ If no access mapping matches the user access token, the user will get an "access
 
 New rules can be added by clicking **Add access mapping** or **Add inventory roles** at the bottom. An access mapping statement can consist of multiple checks like in the image below. You can add a rule to an existing statement by clicking **and**. Click the Minus button to remove a rule.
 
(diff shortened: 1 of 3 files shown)


content/users-guide/administration-bundle/configuring-single-sign-on.md
```diff
@@ -146,7 +146,7 @@ On user login, user data like first name, last name, email and phone number can
 
 Mapping for alias is not available because it is not used in the context of SSO login.
 
-Each access token is signed by a signing certificate. Currently there are three options to configure the signing certificates.
+Each access token is signed by a signing certificate. There are following options to configure the signing certificates.
```
BeateRixen, line 149:
```suggestion
Each access token is signed by a signing certificate. The following options are available to configure the signing certificates.
```

darluk (approved):

BeateRixen (approved):

pawel-rynarzewski-sag (approved):

PR #1051, opened 2023-05-16
PAB-4030: Update info under "Restrictions for output blocks"

kailol-sag:


skom-sag (approved):

content/streaming-analytics/analytics-builder-bundle/wires-and-blocks.md
```diff
@@ -623,8 +623,7 @@ Input and output blocks identify a series of events by specifying a key for the
...
-
-As there can be connections between the models, the main reason for this restriction is to avoid ambiguities or errors that may occur while processing events in the input blocks if there are multiple output blocks \(in different models\) generating the same output stream at the same point in time.
+{{< c8y-admon-important>}}
+In Analytics Builder, for synchronous output types such as measurements, events and alarms \(see also [Output blocks and event timing](/streaming-analytics/analytics-builder/#output-blocks-and-event-timing)\), it is allowed to have more than one output block that generates events with any given key.
+As there can be connections between the models, ambiguities may occur while processing events in the input blocks if there are multiple output blocks (in different models) generating the same output stream at the same point in time. When using output blocks, care must be taken to ensure that no two blocks generate output of the same stream type at the same time.
```
rssag, line 628:
```suggestion
As there can be connections between the models, ambiguities may occur while processing events in the input blocks if there are multiple output blocks (in different models) generating the same output stream at the same point in time. When using output blocks, ensure that no two blocks generate output of the same stream type at the same time.
```

rssag (approved):
a suggestion + approval

PR #1146, opened 2023-06-01
Consistent wording/markup for active/inactive state of Analytics Builder models

kailol-sag:
Trying to be consistent when mentioning active and inactive models. Use lowercase in most cases. And use bold when referring directly to the label that is shown in the UI.


content/streaming-analytics/analytics-builder-bundle/using-the-model-manager.md
```diff
@@ -189,10 +189,10 @@ The inputs that a model receives and what happens to its outputs depends on the
...
 -   **Production**. When active, the model is deployed to the Apama correlator so that the measurements and events from the devices are processed. The output of the model is stored and sent back to the devices.
 
-A model in draft mode can only be in the Inactive state. A model in test, simulation or production mode can be in either the Active or Inactive state.
+A model in draft mode can only be in the **Inactive** state. A model in test, simulation or production mode can be in either the **Active** or **Inactive** state.
```
BeateRixen, line 195: Above, you avoid the phrasing with "state". Not sure, if this is something different here though.

kailol-sag: Yeah, that's tough. Not sure what would be the best. How about this?

```suggestion
A model in draft mode can only be in the inactive state. A model in test, simulation or production mode can be in either the active or inactive state.
```

PR #1158, opened 2023-06-09
MTM-47923-update-for-DM-section

rssag:


content/device-integration.md
```diff
@@ -9,4 +9,4 @@ section:
 weight: 50
 ---
 
-Learn how to set up Mbed u-blox, Raspberry Pi and Tinkerforge as demo devices for proof-of-concept purposes in your Cumulocity IoT environment.
+Set up a Raspberry Pi and Tinkerforge as demo devices in Cumulocity IoT or browse references for *fragments*  used in managed objects (devices) or for device sensors.
```
BeateRixen, line 12: Not sure why fragments is formatted in italics here?

content/thin-edge.md
```diff
@@ -0,0 +1,12 @@
...
+icon: "c8y-icon c8y-icon-device"
+type: root
+layout: redirect
+bundle: thin-edge
+section:
+  - device_management
+weight: 60
+---
+
+Learn more about device integration using thin-edge.io.
```
BeateRixen, line 12: I would add a bit more details here what thin-edge.io is.

content/supported-devices.md
```diff
@@ -0,0 +1,12 @@
...
+icon: "c8y-icon c8y-icon-device"
+type: root
+layout: redirect
+bundle: supported-devices
+section:
+  - device_management
+weight: 70
+---
+
+Supported devices.
```
BeateRixen, line 12: That is not much :-)

content/device-management-application/registering-devices.md
```diff
@@ -0,0 +1,71 @@
+---
+weight: 10
+title: Registering devices
```
BeateRixen, line 3: This is very often referenced throughout the entire documentation in various places (all device protocols, etc.). If we change the name of the file and its title we need to take care of changing all references accordingly, but we anyway change the structure and have to update all links. Just a heads up.

content/device-management-application/managing-device-data-bundle/managing-firmware-on-a-device.md
```diff
@@ -0,0 +1,42 @@
+---
+weight: 20
+title: Managing firmware on a device
```
BeateRixen, line 3: I would not have this as an extra file as this is actually part of the "Managing firmware" section. This should show up as a right-hand heading in there. Same for software.

content/device-management-application/managing-device-data-bundle/applying-device-profiles-to-devices.md
```diff
@@ -0,0 +1,41 @@
+---
+weight: 80
+title: Applying device profiles to devices
```
BeateRixen, line 3: This should not show up as an entry at the left but as a right-hand entry in the Device profiles chapter. Otherwise this is confusing. We have 6 entries in the navigator and currently 9 entries in the index.

BeateRixen:
The Device details section looks weird. Wouldn´t it be clearer if everything is kept in one file and all tabs show up at the right? Othwerwise we have only one entry at the right (Control) because it has sub-headings. I would actually not expect the tabs in the left index. Also, the table can probably be removed at the top?

BeateRixen (approved):

PR #1166, opened 2023-06-14
BugherdTask#715

dianaolives:


MWindel:
These changes only apply to 10.16.0

PR #1170, opened 2023-06-16
Publish the PAQ 10.18 doc with the C8Y 10.17 doc

kailol-sag:
Copied over the entire "streaming-analytics" folder from the 10.18 branch into the 10.17 branch, incl. images.
Removed the "apama" folder, also in static.
Changed the shortcodes to point to PAM 10.15.3.
Updated codeowners file.
Fixed links in other guides.
Verified in Hugo that this looks as expected.


BeateRixen:
@kailol-sag I assume that this PR is on hold until it is 100% clear that Apama 10.18 will be delivered with Cumulocity IoT 10.17? Otherwise we would have to revert at a later point in tome which could cause a lot of conflicts with so many files affected.

kailol-sag:
> @kailol-sag I assume that this PR is on hold until it is 100% clear that Apama 10.18 will be delivered with Cumulocity IoT 10.17? Otherwise we would have to revert at a later point in tome which could cause a lot of conflicts with so many files affected.

Hi @BeateRixen , it is 100% clear that Apama 10.18 is to be delivered with C8Y 10.17. I've only set this to draft mode because it still needs to be reviewed by other dev people.

skom-sag (approved):

ddfourni1 (approved):
Seems correct.

PR #1183, opened 2023-06-28
[DTM] Updated Localization info

lra-sag:
Added note in the Localization section to cover the behaviour for translation of Asset models and names

### DIFF
--- a/content/dtm/localization.md
+++ b/content/dtm/localization.md
@@ -56,6 +56,12 @@ If you want to add translations for the created [asset properties](/dtm/asset-ty
 
 Once you create an asset property or an asset model, their names are added automatically as keys in the **Localization** page. You can further add the translations as required.
 
+{{< c8y-admon-info>}}
+Note that if an asset property or asset model is present, the corresponding entry on the Localization page cannot be deleted. However, if no translation exists for the asset property or asset model, deleting the asset property or asset model will automatically remove the corresponding entry.
+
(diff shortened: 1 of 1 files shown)


pavanshivajirao (approved):

content/dtm/localization.md
```diff
@@ -56,6 +56,12 @@ If you want to add translations for the created [asset properties](/dtm/asset-ty
 
 Once you create an asset property or an asset model, their names are added automatically as keys in the **Localization** page. You can further add the translations as required.
 
+{{< c8y-admon-info>}}
+Note that if an asset property or asset model is present, the corresponding entry on the Localization page cannot be deleted. However, if no translation exists for the asset property or asset model, deleting the asset property or asset model will automatically remove the corresponding entry.
```
MWindel, line 60:
```suggestion
Note that if an asset property or asset model is present, the corresponding entry on the **Localization** page cannot be deleted. However, if no translation exists for the asset property or asset model, deleting the asset property or asset model will automatically remove the corresponding entry.
```

content/dtm/localization.md
```diff
@@ -56,6 +56,12 @@ If you want to add translations for the created [asset properties](/dtm/asset-ty
...
+
+On the other hand, if a translation exists for the asset property or asset model, deleting it will not automatically remove the corresponding entry. In such cases, a delete option will be available for that specific translation, allowing manual deletion if desired.
```
MWindel, line 62:
```suggestion
On the other hand, if a translation exists for the asset property or asset model, deleting it will not automatically remove the corresponding entry. In such cases, a delete option will be available for that specific translation, allowing you to manually delete the entry.
```

PR #1190, opened 2023-07-03
PAB-4144: Make HTTP mechanism clearer

kailol-sag:


content/streaming-analytics/epl-apps-bundle/builtin-actions.md
```diff
@@ -52,6 +52,11 @@ and [Invoking other parts of the Cumulocity IoT REST API]({{< link-apama-webhelp
 
 ### Invoking HTTP services
 
+{{< c8y-admon-info >}}
+The information below is for interacting with *external* HTTP services. For making requests to other parts of the {{< product-c8y-iot >}} REST API, see the information above.
```
bph-sag, line 56: > see the information above.

Is it not plausible to hyperlink to the above header? I get that it is a scroll above, but I've encountered a customer who sent me a hyperlink to this section of the guide, asking how to do the thing right above!

rssag (approved):

content/streaming-analytics/epl-apps-bundle/builtin-actions.md
kailol-sag: Hm, when I copy the "automatic" link from the generated HTML, I get this:
http://localhost:1313/guides/streaming-analytics/epl-apps/#invoking-other-parts-of-the-hahahugoshortcode-s9-hbhb-rest-api
But I do not think that I should use such a link. Will try to add a manual one instead ....

bph-sag: I did notice that as well locally. It's #invoking-other-api on the public documentation, so not sure what magic happened there.

kailol-sag: The public doc is still the one without Analytics Builder. At that time, it was an .md topic of its own. But after the restructuring, it's no longer a separate topic.

kailol-sag: Ok, there's a link now.

bph-sag (approved):

PR #1332, opened 2023-08-22
DataHub-anchors

dianaolives:


content/datahub/working-with-datahub-bundle/configuring-offloading-jobs.md
```diff
@@ -156,7 +153,7 @@ In the **Description** field, you can add a description for this offloading pipe
 
 Click **Next** to proceed with the next configuration step. Click **Previous** to go back one configuration step. Click **Cancel** to cancel the offloading configuration.
 
-<a name="finish"></a>
+<a id="configure-additional-settings"></a>
```
BeateRixen, line 156: The ID notation looks actually different, doesn´t it @rssag?

rssag: it's still an anchor which is to be avoided.

example:
### heading line {#heading-line}

reference:
https://gohugo.io/content-management/cross-references/#heading-ids

BeateRixen (approved):

PR #1375, opened 2023-08-28
MTM-55111 remove misleading statement about clean session which should be always set to true

slc-softwareag:
We always suggest to use `cleanSession=true` so it doesn't make sense to mention about setting it to false. 
Backport should be done to all supported GA versions.


rssag (approved):

rssag:
@slc-softwareag Merging to the develop branch is currently blocked since we rebuild the entire documentation structure. After the rebuild, merging will no longer be possible. You must add your changes manually to the new structure. We communicated this via email some time ago.

slc-softwareag:
@rssag I'm aware of this, what about backports, can we already merge it there?

PR #1379, opened 2023-08-29
Files and shortcodes

BeateRixen:
We still need to adapt the shortcodes for the versions


BeateRixen:
@kailol-sag @sharath-jk @Oezge-Akin Please ignore

BeateRixen:
The preview adminition should not be removed, I will have to ignore this deletion.

MWindel (approved):

PR #1398, opened 2023-09-06
updated compatibility policy and glossary towards yearly releases and CD

BeateRixen:


content/concepts/introduction-bundle/compatibility-policy.md
```diff
@@ -29,7 +29,7 @@ However, {{< product-c8y-iot >}} is continually improving the user experience an
 
 
 {{< c8y-admon-info >}}
-If changes result in breaking backward compatibility, to provide our customers with the time to change their solution, such changes will be announced at least one GA release ahead of the GA release where the change becomes effective. Where the change has a wide-ranging impact, we will delay taking action until we have ascertained that the impact on our customers is minimal.
+If changes result in breaking backward compatibility, to provide our customers with the time to change their solution, such changes will be announced at least 6 months ahead of the point in time, when the change becomes effective. Where the change has a wide-ranging impact, we will delay taking action until we have ascertained that the impact on our customers is minimal.
```
rssag, line 32:
```suggestion
If changes result in breaking backward compatibility, to provide our customers with the time to change their solution, such changes will be announced at least 6 months ahead of when the change becomes effective. Where the change has a wide-ranging impact, we will delay taking action until we have ascertained that the impact on our customers is minimal.
```

+ more occurrences.

content/glossary/c.md
```diff
@@ -6,6 +6,13 @@ section:
   - getting_started
 ---
 
+### Continuous Deployment model {#continuous-deplyoment-model}
+
+In the {{< product-c8y-iot >}} Continuous Deployment (CD) model, the different components of the {{< product-c8y-iot >}} platform are updated on a continuous basis, that is, small incremental improvements are deployed multiple times a week. Continuous updates provide customers with a steady stream of enhancements, bug fixes, and new functionalities as they become available.
```
rssag, line 11:
```suggestion
In the {{< product-c8y-iot >}} Continuous Deployment (CD) model, the different components of the {{< product-c8y-iot >}} platform are updated on a continuous basis, that is, small incremental improvements are deployed multiple times per week. Continuous updates provide customers with a steady stream of enhancements, bug fixes, and new functionalities as they become available. 
```

not wrong, but "per" is more formal (says google :D)

content/glossary/c.md
```diff
@@ -6,6 +6,13 @@ section:
   - getting_started
 ---
 
+### Continuous Deployment model {#continuous-deplyoment-model}
+
+In the {{< product-c8y-iot >}} Continuous Deployment (CD) model, the different components of the {{< product-c8y-iot >}} platform are updated on a continuous basis, that is, small incremental improvements are deployed multiple times a week. Continuous updates provide customers with a steady stream of enhancements, bug fixes, and new functionalities as they become available. 
+
+See also [Private Preview](/glossary/p/#private preview), [Public Preview](/glossary/p/#public preview) and [General Availability](/glossary/g/#ga).
```
rssag, line 13: the anchor IDs don't look correct

edit: checked locally. the links fully break. "_" should be included.
i'd advise using "-" instead of "_" for the IDs though.

rssag (changes requested):

content/concepts/introduction-bundle/compatibility-policy.md
BeateRixen: I also do not like my phrasing but are you sure, your suggestion is correct? For me it sounds wrong :-). How about "ahead of the change becoming effective". I was thinking about that one earlier? What do you think, @MWindel ?

content/glossary/c.md
BeateRixen: I don´t get what you mean. What is the difference to the IDs we already have in place in the glossary?

rssag: the references break, because there's a space in the URL

updated the comment, the underscore did not show. i'd advise using hyphen instead of underscore in IDs.

BeateRixen: You are right, no idea, why I used underscores :-). Replaced them.

content/concepts/introduction-bundle/compatibility-policy.md
MWindel: @BeateRixen @rssag I like both versions. The gerund is a bit smoother in my opinion.

rssag: googled specifically for "ahead of when". seems to be common, doesn't sound wrong to me

BeateRixen: Just changed it now to the gerund :-). But if prefer the other way, I can again change it. This will be reviewed by Niko and Michael as this is of legal relevance. This is the reason why I was so precise with my phrasing. Let´s see what they suggest.

MWindel (approved):

PR #1407, opened 2023-09-07
MTM-54143: reference-notifications2: type-filtering-changes-consistency and improve text on contexts and MO lifecycle event messages

gapa-c8y:
Two small changes:
 - move (and re-contextualize) the sentence about tenant context alarms and events from between the 2 sentences about MO lifecycle-boundary events (creates and deletes), so those sentences are together.
 - add single quotes (as default approach) to filterType example (consistent with openapi docs).

This should be backported to 1017 and 1016 and will need porting to the 1019/develop location for this documentation (expected to be openapi doc, but not done yet)


sagscmi (approved):

content/reference/notifications-bundle/overview.md
```diff
@@ -211,11 +211,13 @@ Therefore, to receive notifications informing of new managed object creations, c
 to listen for them.
 An application can use this to discover new managed objects.
 It can then choose to create subscriptions with "mo" **context** for those managed objects.
-It is also possible to subscribe to all alarms or events that are generated in a tenant using "tenant" **context**.
 
 Notifications for managed object deletions are forwarded to topics by subscriptions with "mo" (managed object) **context** 
 that include the "managedobjects" API, and by subscriptions with "tenant" **context** that include the "managedobjects" API.
 
+Subscriptions with "tenant" **context** can also use the alarms API and/or the events API to forward all alarms and/or events,
+respectively, that occur within the tenant, applying filters as desired.
```
rssag, line 219:
```suggestion
Subscriptions with "tenant" context can also use the alarms API and/or the events API to forward all alarms and/or events,
respectively, which occur within the tenant, applying filters as desired.
```

content/reference/notifications-bundle/overview.md
```diff
@@ -286,7 +288,7 @@ To include messages of **type** "temperature" only:
     "id": 2468
   },
   "filter": {
-    "type": "temperature"
+    "type": "'temperature'"
```
rssag, line 291: out of curiosity - why the double quotes?

rssag:
will there be a separate PR for develop/2024?

content/reference/notifications-bundle/overview.md
gapa-c8y: @rssag, 'context' in this line (and in most other usages in this area) is referring to the context JSON field, which I why it is in bold in all those places. It is the field we are referring to here also, which is why I bold formatted it. 

If you think we are miss-using the bold formatting, maybe we should review other cases also.

What's your thoughts on this?

content/reference/notifications-bundle/overview.md
gapa-c8y: We wish to provide a string with quotes (hence we nest single quotes in double quotes).
Though, when specifying one type value, the behaviour is the same whether the single-quotes are included or not, PO Scott wanted us to promote usage of quotes as that is a more general approach that can be used to provide multiple type values here.
For example, to match temperature or pressure (either, as the messages this will try to match can have one type only), we would use the field value "'temperature' OR 'pressure'" (the quotes distinguish this from a single type value of "temperature OR pressure").

This is consistent with the change we are doing for the openapi in MTM-54143 (https://github.softwareag.com/IOTA/cumulocity-openapi/pull/196), but here the case is s simple single item to (possibly) match in a message.

Hope that makes sense!

gapa-c8y:
> will there be a separate PR for develop/2024?

The file we are updating (for reference/Notification 2.0) is expected to be moved to the openapi doc, as in a discussion You, Beate, Scott and I had by email (last message on 2023-08-23, title 'Re: Notifications section in documentation').

This move is something I wanted to catch up with you; were you offering to do this move, or should we prepare a PR for it.

We can apply this change before or after the move, I was assuming after. 

Again, let me know you thoughts/preference here.

content/reference/notifications-bundle/overview.md
rssag: no, if it was a deliberate choice, i'm ok with it :)
what about "that" vs "which"? (it's not too important, i'd be fine with both)

content/reference/notifications-bundle/overview.md
rssag: ok, thanks :)

content/reference/notifications-bundle/overview.md
gapa-c8y: > what about "that" vs "which"? (it's not too important, i'd be fine with both)

I'll correct it - but am a bit busy handling some customer issues, so this might not move for a day or so

rssag:
notifications 2.0 for openapi doc PR here:
https://github.softwareag.com/IOTA/cumulocity-openapi/pull/201

i added the changes from the c8y-docs PR there too.

gapa-c8y:
@rssag, I've just accepted the last change suggestion, so this is ready as far I'm concerned. Thanks for including this in the openapi PR also. That last sugggested change (from you, changing a 'that' to a 'which') is not in the openapi version of this, but you seemed quite relaxed about the change (so up to you if that is also transferred to openapi, I guess).

Let me know if any other changes needed and if it can be automatically backported (or done with minimal effort) otherwise I can do so.
Thanks

rssag:
since it's not targeting develop, i'll cherry-pick this manually.
can i merge?

gapa-c8y:
> since it's not targeting develop, i'll cherry-pick this manually. can i merge?

Please go ahead - thanks

slc-softwareag (approved):

PR #1421, opened 2023-09-13
added audience labels

BeateRixen:


BeateRixen:
Not 100% sure in all cases. But I like it, I think it is useful. I asked Carlos to create more, that is, Administrator and Business User. If you have more ideas, let him know.

rssag (approved):

PR #1422, opened 2023-09-13
Adapt short codes

BeateRixen:


BeateRixen:
I have kept the short code "current-version" for now (and set it to 10.18) as we have a couple of occurrences under discussion at the moment and I don´t want to break the build.

BeateRixen:
- Removed the "current-version" short code from all API links. 
- Removed several short codes no longer in use.
- Rephrased several sections towards CD.

kailol-sag (approved):
I only reviewed the Streaming Analytics stuff, and this looks ok to me. Thanks :-)

Oezge-Akin (approved):

themes/c8ydocs/layouts/shortcodes/c8y-current-version.html
```diff
@@ -1 +1 @@
-{{- "10.19.0" -}}
+{{- "10.18.0" -}}
```
rssag, line 1: so we do keep the shortcode? :)

rssag (approved):

themes/c8ydocs/layouts/shortcodes/c8y-current-version.html
BeateRixen: Yes, as mentioned above in my comment I keep it for now (and set it to 10.18) as we have a couple of occurrences under discussion at the moment and I don´t want to break the build.

PR #1423, opened 2023-09-14
Add missing slash to API doc links

kailol-sag:


BeateRixen (approved):

kailol-sag:
> @kailol-sag Does that have to be fixed in older release branches as well?

Jep, it's also wrong in 10.17 and 10.18.
For Streaming Analytics, the folder structure is the same in the old versions and in CD. So would it be possible to backport or cherry-pick this? Or do I have to create a new PR?

MWindel (approved):

PR #1424, opened 2023-09-14
[MTM-55028] Document that SSO in c8y does not support SAML based auth…

Michal-Bak:
Document that SSO in c8y does not support SAML-based authentication, only OpenID and OAuth2

### DIFF
--- a/content/authentication/sso-bundle/integration-with-azure.md
+++ b/content/authentication/sso-bundle/integration-with-azure.md
@@ -7,7 +7,7 @@ section:
 ---
 
 
-The integration was successfully verified against Azure AD. The configuration steps are available in [https://docs.microsoft.com/en-us/azure/active-directory/develop/v1-protocols-oauth-code](https://docs.microsoft.com/en-us/azure/active-directory/develop/v1-protocols-oauth-code).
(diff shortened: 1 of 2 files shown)


BeateRixen (approved):

kowalczykmike (approved):

BeateRixen:
@Michal-Bak For which versions is that change relevant? Only after 10.18? Otherwise we need a second PR for 10.18 and lower versions.

Michal-Bak:
> @Michal-Bak For which versions is that change relevant? Only after 10.18? Otherwise we need a second PR for 10.18 and lower versions.

It can also go for 10.18 and lower. Should I open a new PR to the "release/r10.18.0" branch?

rssag:
> > @Michal-Bak For which versions is that change relevant? Only after 10.18? Otherwise we need a second PR for 10.18 and lower versions.
> 
> It can also go for 10.18 and lower. Should I open a new PR to the "release/r10.18.0" branch?

note that the doc structure changed significantly. we cannot directly cherry-pick to 10.18.0 and earlier, a new PR to release/r10.18.0 is the best approach, yes (we can cherry-pick to earlier releases from there).

PR #1437, opened 2023-09-19
cherry-pick MTM-54143 to 10.16.0

rssag:


rssag:
@gapa-c8y

i created this because i'm not sure how to resolve the conflict or how it occurred.

gapa-c8y:
> @gapa-c8y
> 
> i created this because i'm not sure how to resolve the conflict or how it occurred.

I'll take a look later (meetings all this morning)

gapa-c8y:
> @gapa-c8y
> 
> i created this because i'm not sure how to resolve the conflict or how it occurred.

I think I may have incorrectly assumed (or thought) that the major change we made to the reference/Notification-2 section of the user-docs had gone back to 1016. It appears it hasn't and hence the conflict. 

So the issue is (for me) should it have gone to 1016. I'll check with PO, but I suspect we might NOT do this. I'll update once I confirm

gapa-c8y:
> > @gapa-c8y
> > i created this because i'm not sure how to resolve the conflict or how it occurred.
> 
> I think I may have incorrectly assumed (or thought) that the major change we made to the reference/Notification-2 section of the user-docs had gone back to 1016. It appears it hasn't and hence the conflict.
> 
> So the issue is (for me) should it have gone to 1016. I'll check with PO, but I suspect we might NOT do this. I'll update once I confirm

@rssag - I've confirmed with PO that this is okay. That is, we do not need to backport this (or any earlier changes) to 1016. Sp this PR can be closed and not done.

I'll close now

rssag:
> @rssag - I've confirmed with PO that this is okay. That is, we do not need to backport this (or any earlier changes) to 1016. So this PR can be closed and not done.
> 
> I'll close now

thanks for the effort :)

PR #1444, opened 2023-09-20
MTM-55444: edge-databroker: add new uninstall steps file (develop)

gapa-c8y:
This is the develop version of this 1018 PR: https://github.com/SoftwareAG/c8y-docs/pull/1443 

Updated to match content that was merged to the 1018 trunk

I'm not certain of the added page's 'layout' and 'section' settings (at the top of the page source). I have just copied them from the sibling (install) page.


content/edge/edge-databroker-bundle/databroker-edge-uninstall.md
```diff
@@ -0,0 +1,149 @@
...
+```
+6. Delete each of the persistent volumes found in the previous step.
+```shell 
+sudo kubectl delete pv journal0 ledgers0 zookeeper0
+```
+7. Delete the directories used by the persistent volumes.
+```shell
+sudo rm -rf /opt/bookie/ledgers /opt/bookie/journal /opt/zookeeper
+```
+8. Unsubscribe the edge tenant from and delete, the databroker-agent-server microservice.
```
rssag, line 48: something seems to be missing here

content/edge/edge-databroker-bundle/databroker-edge-uninstall.md
```diff
@@ -0,0 +1,149 @@
...
+   To find its application ID, you can use a cURL command of the following form and redirect the output to a jq command to make it easier to read the JSON response.
+   The general structure of the command is:
+```shell
+curl http://<EDGE-HOSTNAME>/application/applicationsByName/databroker-agent-server -su <TENANT>/<USERNAME> | jq --indent 2
+```   
+   The ID required is the value of the `id` field at the root level of the JSON response.
+   In the following example, it has the value '19' (the penultimate JSON field).
+```shell
+curl http://myown.iot.com/application/applicationsByName/databroker-agent-server -su edge/admin | jq --indent 2
+Enter host password for user 'edge/admin':
```
rssag, line 60: this line is currently inside the shell code block. shouldnt we let that block end before this line and open another code block after this line?

content/edge/edge-databroker-bundle/databroker-edge-uninstall.md
```diff
@@ -0,0 +1,149 @@
...
+   Use that ID value in the request to unsubscribe from the microservice and again when deleting it. 
+
+   To unsubscribe, use a command of the following form:
+```shell
+curl -X DELETE http://<EDGE-HOSTNAME>/tenant/tenants/<TENANT>/applications/<APP-ID> -u <TENANT>/<USERNAME> -v
+```
+   This returns an HTTP 204 (no content) response. For example: 
+```shell
+curl -X DELETE http://myown.iot.com/tenant/tenants/edge/applications/19 -u edge/admin -v
+Enter host password for user 'edge/admin':
```
rssag, line 95: same here, i think this line should not be part of a code block

content/edge/edge-databroker-bundle/databroker-edge-uninstall.md
gapa-c8y: The line "Enter host password for user 'edge/admin':" comes from the command (it prompts the user for their password). The password is not echoed (shown)as it is typed and, assuming it is correct/successful, the response is returned.|

So all of this is the output from the (one) curl command. So I think this should remain as one code block

content/edge/edge-databroker-bundle/databroker-edge-uninstall.md
gapa-c8y: As above (and one more instance below)

content/edge/edge-databroker-bundle/databroker-edge-uninstall.md
gapa-c8y: It's the same as the original PR and as I intended. 

The numbered line/point declares what we are about to do. These are grouped together  as they both depend on getting the application ID from the server first, so we do that first then the 2 requests that need it.

If you think we should structure it differently, I'm happy to change it. But as stated, it appears to be as I intended

content/edge/edge-databroker-bundle/databroker-edge-uninstall.md
rssag: alright

slc-softwareag (approved):

PR #1497, opened 2023-10-18
MTM-56190 Rename MQTT Connect to MQTT Service

slc-softwareag:


gapa-c8y (approved):

gapa-c8y:
A separate change: I think the 'list' of arrow colors in the diagram should be labeled as 'Key' or 'Legend' and put in a black border away from other components. At a glance it looks like 3 things going into the messaging-service component

rssag (approved):

PR #1539, opened 2023-11-16
BugherdTask#760

dianaolives:


BeateRixen:
@dianaolives Is there a specific reason why you renamed this icon?

BeateRixen:
@dianaolives One more question :-). The size is not the same as before. Is this the original size?

dianaolives:
> @dianaolives One more question :-). The size is not the same as before. Is this the original size?

I made it the same size as the search button 30x27. Just because since the new icon has a letter on it, in the original size it looked too small.

BeateRixen (approved):

MWindel (approved):

rssag (approved):

PR #1545, opened 2023-11-17
PAB-4296: Diagnostics links are now in the right drawer

kailol-sag:


skom-sag (approved):

kailol-sag:
Nabu confirmed on a chat that all is good. Therefore,merging this now.

PR #1552, opened 2023-11-22
Web sdk tutorials update

MWindel:
- Added missing links
- Rephrased existing links
- Added additional information for increased clarification
- replaced "need to" with must or other equivalent in accordance to our Style guide


sag-lukjan (approved):

content/web/appendix-migration-history.md
```diff
@@ -177,7 +177,7 @@ The {{< product-c8y-iot >}} platform itself is running as a hybrid application i
 
 #### Deprecating custom manifests {#deprecating-custom-manifests}
 
-Now developers have at their disposal all modern JS, so they just need to use normal ESM to import and export dependencies. Although the manifest files will still work, they are no longer required. We still support manifest files as entry points, so it is possible to use these as well as normal package.json files.
+Now developers have at their disposal all modern JS, so they must use normal ESM to import and export dependencies. Although the manifest files will still work, they are no longer required. We still support manifest files as entry points, so it is possible to use these as well as normal package.json files.
```
BeateRixen, line 180: probably better "just have to"? Otherwise it is a different meaning.

content/web/libraries-bundle/client-library.md
```diff
@@ -78,7 +78,7 @@ In the following sections, the default signature of these functions is described
 
 ### Accessing a microservice with the Fetch API {#accessing-a-microservice-with-the-fetch-api}
 
-The client internally uses the Fetch API. By accessing this core function, you can do any authenticated request to any resource. Standalone you can use `core.client.fetch(url, options)` and in `@c8y/ngx-components/data` for Angular you simply need to inject the `FetchClient`:
+The client internally uses the Fetch API. By accessing this core function, you can do any authenticated request to any resource. Standalone you can use `core.client.fetch(url, options)` and in `@c8y/ngx-components/data` for Angular you must inject the `FetchClient`:
```
BeateRixen, line 81: Same here. Probably better "simply have to". Otherwise the meaning changes. Below, you say "you simply must". Sounds somehow strange to me :-), but if you say that this is correct, would work here as well.

BeateRixen:
@MWindel This PR will conflict with https://github.com/SoftwareAG/c8y-docs/pull/1541.

content/web/appendix-migration-history.md
MWindel:
```suggestion
Now developers have at their disposal all modern JS, so they just have to use normal ESM to import and export dependencies. Although the manifest files will still work, they are no longer required. We still support manifest files as entry points, so it is possible to use these as well as normal package.json files.
```

content/web/libraries-bundle/client-library.md
MWindel:
```suggestion
The client internally uses the Fetch API. By accessing this core function, you can do any authenticated request to any resource. Standalone you can use `core.client.fetch(url, options)` and in `@c8y/ngx-components/data` for Angular you simply have to inject the `FetchClient`:
```

PR #1557, opened 2023-11-23
PAQ: Moved info on audit logs

kailol-sag:
The audit logs are for the entire Streaming Analytics application, not only for Analytics Builder.
Therefore moved this info into the troubleshooting section, and rewrote it so that it also mentions EPL apps.
Also updated/reworded the link in the standard-tenant section so that it now points to the new topic.


content/streaming-analytics/troubleshooting-bundle/analytics-audit-logs.md
```diff
@@ -0,0 +1,28 @@
+---
+weight: 30
+title: Audit logs for Streaming Analytics
+layout: redirect
+---
+
```
kailol-sag, line 6: Many topics in "Troubleshooting and diagnostics" say that this "is not available for the Apama-ctrl-smartrules and Apama-ctrl-smartrulesmt microservices." Do we also have to add this Info note to this new topic?

content/streaming-analytics/troubleshooting-bundle/analytics-audit-logs.md
```diff
@@ -0,0 +1,28 @@
...
+Activation and deactivation of analytic models and EPL apps is shown in the audit logs. The audit logs are accessible via the Administration application and the audit API.
+See [Audit logs](/standard-tenant/audit-logs/) and [Audit API](https://{{< domain-c8y >}}/api/core/#tag/Audit-API)
+in the {{< openapi >}} for details of accessing the audit logs.
+
+Audit log entries include the current action and the the name of the user performing that action.
+The following is an example of an audit log entry for an analytic model that runs in simulation mode (with additional line breaks in the `text` field for better readability):
```
rthrippleton, line 12: I've always felt that this example here seems too technical/unfriendly - the common user uses the UI, not the raw API. The core admin doc uses screenshots as audit examples.

Would you be interested in a screenshot to paste here instead? (I can easily get one)

content/streaming-analytics/troubleshooting-bundle/analytics-audit-logs.md
rthrippleton: Probably not. It's true, but it is _implicitly_ true rather than a limitation of those microservices, because of the features they support.

That is, we have audit logging for Analytics Builder and EPL operations, but those microservices do not support Analytics Builder and EPL, therefore no audit logs.

content/streaming-analytics/troubleshooting-bundle/analytics-audit-logs.md
kailol-sag: We try to  use as few screenshots as possible - so we do not have to update them too often. But if you think it helps ... let me see how such a screenshot would look like.

rthrippleton: audit
Feel free to clip it as close as you'd like.

kailol-sag: Thanks Richard. I've updated the doc, and I got rid of the mouse pointer in your example ;-)
Better now?

content/streaming-analytics/troubleshooting-bundle/analytics-audit-logs.md
```diff
@@ -0,0 +1,14 @@
...
+weight: 30
+title: Audit logs for Streaming Analytics
+layout: redirect
+---
+
+Activation and deactivation of analytic models and EPL apps is shown in the audit logs. The audit logs are accessible via the Administration application and the audit API.
+See [Audit logs](/standard-tenant/audit-logs/) and [Audit API](https://{{< domain-c8y >}}/api/core/#tag/Audit-API)
+in the {{< openapi >}} for details of accessing the audit logs.
+
+Audit log entries include the current action and the the name of the user performing that action.
```
BeateRixen, line 11:
```suggestion
Audit log entries include the current action and the name of the user performing that action.
```

BeateRixen (approved):

rthrippleton (approved):

PR #1579, opened 2023-12-07
bugherd task 759 - section restored for 1018

rssag:


BeateRixen:
@rssag I just read your comment in the BH ticket that this section was restored from 10.10. So it was not removed intentionally when you did the rebuild? Is the content still true :-)?

MWindel (approved):

PR #1585, opened 2023-12-07
MTM-55819 changing TS parameters description to clear inconsistency

deiv3:


BeateRixen:
@deiv3 This change is for 10.18 but the text is talking about 10.16 which is confusing.

BeateRixen (approved):

jaro-herod (approved):

PR #1591, opened 2023-12-14
MTM-34284_supportUrl_updated_documentation

darluk:


karolkaczmarski-sag (approved):

jkot-softwareag (approved):

pmich-sag (approved):

content/enterprise-tenant/customization-bundle/configuration.md
```diff
@@ -53,9 +53,7 @@ Under **Two-factor authentication**, you can change the SMS template which is se
 
 ### Support link {#support-link}
 
-In the **Support link** section, you can enter a URL to be used to link to a support page. If you do not provide a link here, the default link to the {{< sag-dev-community >}} page will be used.
-
-Enter "false" to hide the link.
+In the **Support link** section, you can enter a default URL to be used to link to a support page (available in "Request support" option in the top-right user menu). If you leave it empty or set it to "false", then the link will not be shown by default. However, an application can override this setting by defining "supportUrl" application option.
```
BeateRixen, line 56:
```suggestion
In the **Support link** section, you can enter a URL which is linked to by default if users click **Request support** in the user menu.  If you leave this field empty or set it to "false", then the link will not be shown by default. However, an application can override this setting by defining the "supportUrl" application option.
```

BeateRixen: then the link will not be shown by default > what does that mean? Shown where? And what happens in this case if I click Request support? Or is this option not available in this case?

darluk: when value is set to 'false', the `request support` is not shown in the user menu at all, once user set the URL it is shown in the user menu

content/enterprise-tenant/customization-bundle/configuration.md
```diff
@@ -53,9 +53,7 @@ Under **Two-factor authentication**, you can change the SMS template which is se
 
 ### Support link {#support-link}
 
-In the **Support link** section, you can enter a URL to be used to link to a support page. If you do not provide a link here, the default link to the {{< sag-dev-community >}} page will be used.
-
-Enter "false" to hide the link.
+In the **Support link** section, you can enter a URL which is available under **Request support** option in the user menu. If you set it to "false", then **Request support** option in the user menu will not be shown. If you leave **Support link** section empty, URl will be taken from tenant options. However, an application can override this setting by defining the "supportUrl" application option.
```
BeateRixen, line 56:
```suggestion
In the **Support link** section, you can provide a URL which is then used as default link for the **Request support** option in the user menu. If you set this value to "false", then the **Request support** option in the user menu will be hidden. If you leave the **Support link** field empty, the URL for the link will be taken from the tenant options. However, an application can override this setting by defining the "supportUrl" application option.
```

BeateRixen:
@darluk After reading your comment I had to update it once more.

BeateRixen (approved):

content/enterprise-tenant/customization-bundle/configuration.md
```diff
@@ -53,9 +53,7 @@ Under **Two-factor authentication**, you can change the SMS template which is se
 
 ### Support link {#support-link}
 
-In the **Support link** section, you can enter a URL to be used to link to a support page. If you do not provide a link here, the default link to the {{< sag-dev-community >}} page will be used.
-
-Enter "false" to hide the link.
+In the **Support link** section, you can provide a URL which is then used as default link for the **Request support** option in the user menu. If you set this value to "false", then the **Request support** option in the user menu will be hidden. If you leave the **Support link** field empty, the URL for the link will be taken from the tenant options. However, an application can override this setting by defining the "supportUrl" application option.
```
pawel-rynarzewski-sag, line 56: I know it's merged already, but wondering if we should precise "tenant options", because this setting also sets one kind of tenant option (the one in the current tenant)? Like "from the inherited tenant options or from the system options." and add a sentence at the end: "If none of the options is set, the setting defaults to "false"".

darluk: when the value is false, the link will be hidden, and `false` is not a default value

pawel-rynarzewski-sag: `false` is the default when none of the options is set: https://github.softwareag.com/IOTA/cumulocity-ui/pull/4991/files#diff-4e32cbaa3780659d5d3d3fad4aa5cc3bb0c6f346b191cdb79da15ef5992da79bR110

PR #1665, opened 2024-02-05
Stop using term 'real-time engine'

rthrippleton:
I think it's a legacy from back when this document covered both Esper and Apama. 'streaming analytics engine' is more specific, and is used earlier in https://cumulocity.com/docs/concepts/realtime/#realtime-processing-in-cumulocity

### DIFF
--- a/content/concepts/realtime-bundle/implementation.md
+++ b/content/concepts/realtime-bundle/implementation.md
@@ -6,9 +6,9 @@ weight: 40
 
 {{< product-c8y-iot >}} provides several processing modes for API requests:
 
-* **persistent** - Stores data in the {{< product-c8y-iot >}} database and sends data to the streaming analytics engine. Afterwards, {{< product-c8y-iot >}} returns the result of the request. This is the default mode.
+* **persistent** - Stores data in the {{< product-c8y-iot >}} database and sends data to the Streaming Analytics engine. Afterwards, {{< product-c8y-iot >}} returns the result of the request. This is the default mode.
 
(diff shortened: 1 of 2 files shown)


content/concepts/realtime-bundle/implementation.md
```diff
@@ -6,9 +6,9 @@ weight: 40
 
 {{< product-c8y-iot >}} provides several processing modes for API requests:
 
-* **persistent** - Stores data in the {{< product-c8y-iot >}} database and sends data to the real-time engine. Afterwards, {{< product-c8y-iot >}} returns the result of the request. This is the default mode.
+* **persistent** - Stores data in the {{< product-c8y-iot >}} database and sends data to the streaming analytics engine. Afterwards, {{< product-c8y-iot >}} returns the result of the request. This is the default mode.
```
kailol-sag, line 9: Confused about the case here - why no initial caps? You said that earlier in the doc it is named "Apama streaming analytics engine". Therefore, I'm assuming that "streaming analytics" is not the product name but just the general term for that kind of functionality. So if you want to keep the lowercase, then I suggest to also add "Apama" here, and also in the changed sentence below.

Also: what exactly is the engine? Is this the Apama-ctrl microservice or something else?

rthrippleton: That's a good point actually - 'streaming analytics' is exclusively a product name, not a generic term, so should be capitalized earlier in the doc as well. I've pushed that change, does this all appear more consistent now?

> Also: what exactly is the engine? Is this the Apama-ctrl microservice or something else?
It's the apama-ctrl microservice. I think 'real-time engine' was used historically to cover both Esper and Apama-ctrl.

kailol-sag (approved):

BeateRixen (approved):

PR #1681, opened 2024-02-07
PAQ 24.41.4 release note for security vulnerability (PAB-4378)

kailol-sag:
Copied over change log entry from develop and adapted the PAQ build number, ticket number, and Microservice SDK version.

### DIFF
--- a/content/change-logs/analytics/apama-in-c8y-24-41-4-Updated-Microservice-SDK-for-vulnerable-third-party-components.md
+++ b/content/change-logs/analytics/apama-in-c8y-24-41-4-Updated-Microservice-SDK-for-vulnerable-third-party-components.md
@@ -1,0 +1,17 @@
+---
+date: 2024-02-07
+title: Updated Microservice SDK for vulnerable third-party components
+product_area: Analytics
+change_type:
+  - value: change-2c7RdTdXo4
+    label: Improvement
+component:
+  - value: component-M5-cepIIS
+    label: Streaming Analytics
+build_artifact:
+  - value: tc-KXXmo2SUR
+    label: apama-in-c8y
+ticket: PAB-4378
+version: 24.41.4
+---
+To fix a security vulnerability, the Microservice SDK has been updated to version 1018.540.31 to use version 1.2.13 of *logback-core* and *logback-classic*.


skom-sag:
should we call this a Fix instead of an Improvement on the 2024 line ?

kailol-sag:
> should we call this a Fix instead of an Improvement on the 2024 line ?

Ok, I can change his.
But I would also like to hear what Beate thinks. 

We had this other PR for the CD line which was just merged into develop, and which has the "Improvement" label so that it will show up in the change log (keep in mind that fixes are not to be shown in the change log). Would you later (in a about year when the branch for the 2025 yearly release is created) also want to change this to "Fix" on the 2025 branch? This would require manual work - and that we do not forget about this! 

Or is there another way to do this? (A "Security" label as indicated on the iTrac does not sound like a bad idea to me.)

BeateRixen:
> > > should we call this a Fix instead of an Improvement on the 2024 line ?
> > 
> > 
> > Ok, I can change his. But I would also like to hear what Beate thinks.
> > We had this other PR for the CD line which was just merged into develop, and which has the "Improvement" label so that it will show up in the change log (keep in mind that fixes are not to be shown in the change log). Would you later (in a about year when the branch for the 2025 yearly release is created) also want to change this to "Fix" on the 2025 branch? This would require manual work - and that we do not forget about this!
> > Or is there another way to do this? (A "Security" label as indicated on the iTrac does not sound like a bad idea to me.)
> 
> I do not want to change Improvement to Fix once 2025 branch is created, lets not do any additional work here.
> 
> Security is an important topic and would be good if we can improve the doc tooling to include this new section, but I would leave it to @BeateRixen. Perhaps you can discuss about adding a new security label in the workstream ?

@kailol-sag See Nikos reply in the release notes data document. I would also not want to add more change types for now but see how we come along with the existing ones after we got used to them.

But I agree I would use the same change type for the yearly release and for CD.

BeateRixen (approved):

kailol-sag:
> But I agree I would use the same change type for the yearly release and for CD.
Ok, I'll keep the "Improvement" for consistency.

PR #1682, opened 2024-02-08
[MTM-56417] Cleanup managing applications doc.

eniosultan:


BeateRixen:
@eniosultan It would be cleaner to change the anchor name and the label for the context help to "packages" (also in the json file), but I leave this up to you. If you remember why it says "extensions" but shows the packages content, I can live with it :-).

BeateRixen (approved):

jdreSAG (approved):

PR #1689, opened 2024-02-09
Create cumulocity-10-20-131-0-modified-behavior-on-creating-an-extern…

BeateRixen:
…al-id-without-an-existing-associated-global-id.md


BeateRixen:
@martin-sag Please especially check if the title and the change type match the ticket description.

martin-sag (approved):

MWindel (approved):

PR #1698, opened 2024-02-13
PAB-4302 Cumulocity microservices can use kubernetes shutdown hooks

shaaf-sag:
Documentation for Cumulocity microservices can use kubernetes prestop hook.

### DIFF
--- a/content/microservice-sdk/general-aspects-bundle/microservice-manifest.md
+++ b/content/microservice-sdk/general-aspects-bundle/microservice-manifest.md
@@ -49,6 +49,14 @@ Here is an example manifest:
         "initialDelaySeconds": 20,
         "periodSeconds": 10
     },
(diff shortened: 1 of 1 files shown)


rthrippleton (changes requested):
From the story:

> 
Very minimal doc to avoid curious users stumbling across this and wanting support, to say that anything in the microservice manifest not explicitly documented is not supported (not mentioning these features by name at all!)

So we actually want a note that's more along the lines of some internal features of the manifest are used by our components, but they are not support for customer operations. Nothing hinting at what they are! It's just to warn off people who are more intimately familiar with the platform, and try to use features that they 'discover'. That then gives us the freedom to change what we've implemented, without it breaking customer solutions.

rthrippleton:
#1706

PR #1703, opened 2024-02-13
CDH-4204 DataHub on Edge on Kubernetes

UlrichPost:
This is really a very first draft requiring initial review by 
* Christoph Heinz
* Tim Doernemann
* Bhaskar Reddy Byreddy

Several open questions have been marked by "???"


bbyreddy (commented):
@UlrichPost , added some comments based on my understanding.

- [ ] In general, we need to keep the Namespace `c8yedge` as default everywehre and put one INFO statement to mention that the commands and the manifests have to be updated with the right namespace name.
- [ ] All the shell commands can be specified in the shell box (as the latest doc allways user to copy the contents with a single click)
- [ ] We can have the manifest files as attachments instead of inlining them in the doc.

PR #1705, opened 2024-02-15
Fix broken links

cheinz-sag:
Hi,

can you pls review and merge to develop? Pls check for first link if using query params in url string is ok for you.

Thanks
Christoph


BeateRixen (approved):

BeateRixen:
> Hi,
> 
> can you pls review and merge to develop? Pls check for first link if using query params in url string is ok for you.
> 
> Thanks Christoph

Yes, that is exactly why we have added them :-).

PR #1709, opened 2024-02-16
Remove the SDK update changelog entry

rthrippleton:
This is far too 'routine' to merit an entry (agreed by Kev)


BeateRixen (approved):

kailol-sag:
Arrggg, before I'm asked to add something to the change log or release notes, can you please ALL agree on whether an entry is required or not?

And you don't even want to list this as a Fix in the yearly release?

kailol-sag (approved):

PR #1794, opened 2024-03-07
PAM-34632: Copied over PAM release note for wew utility functions for the TimeFormat event library

kailol-sag:
The original release note is at https://docbuilder1.eur.ad.sag/jenkins/job/pam/job/trunk-webhelp_en/lastSuccessfulBuild/artifact/out/webhelp/index.html#page/pam-webhelp%2Fco-ApaRelNot_10155_epl_enhancements.html

@Rob-J-SAG , I've added "EPL" to the title - is this enough for the user to understand that this is about Streaming Analytics?

Also not sure about the reference to the ApamaDoc. As the doc for PAM 1015.5 is not yet live, I cannot add a link. Biut is it possible for the user to see the updated ApamaDoc somewhere? Or should we remove the reference to the ApamaDoc?

### DIFF
--- a/content/change-logs/analytics/apama-in-c8y-20240307-New-utility-functions-for-the-EPL-TimeFormat-event-library.md
+++ b/content/change-logs/analytics/apama-in-c8y-20240307-New-utility-functions-for-the-EPL-TimeFormat-event-library.md
@@ -1,0 +1,25 @@
+---
+date: 
+title: New utility functions for the EPL TimeFormat event library
+change_type:
+  - value: change-2c7RdTdXo4
+    label: Improvement
+product_area: Analytics
+component:
+  - value: component-M5-cepIIS
+    label: Streaming Analytics
+build_artifact:
+  - value: tc-KXXmo2SUR
+    label: apama-in-c8y
+ticket: PAM-34632
+version:
+---
+The following utility functions have been added to the TimeFormat event library to help with comparing and manipulating datetimes. As with the existing functions, the new utility functions work for the local time zone, an arbitrary time zone and, where appropriate, the UTC time zone.
+
+* `dateComponent`: Gives the datetime for the previous midnight in the time zone at the given datetime.
+* `timeComponent`: Gives the number of seconds since the previous midnight in the time zone at the given datetime.
+* `daysSinceEpoch`: Gives the number of Julian days since the beginning of the epoch at the given datetime.
+* `getOffset`: Gives the offset in seconds of the time zone from UTC at the given datetime, taking into account any daylight savings that may be being applied.
+* `getRawOffset`: Gives the base offset in seconds of the time zone, that is, without daylight savings applied.
+
+For usage information, see the API Reference for EPL (ApamaDoc).


nmfa (approved):
Looks consistent with PAM version.

Rob-J-SAG (approved):

content/change-logs/analytics/apama-in-c8y-20240307-New-utility-functions-for-the-EPL-TimeFormat-event-library.md
```diff
@@ -0,0 +1,25 @@
...
+---
+The following utility functions have been added to the TimeFormat event library to help with comparing and manipulating datetimes. As with the existing functions, the new utility functions work for the local time zone, an arbitrary time zone and, where appropriate, the UTC time zone.
+
+* `dateComponent`: Gives the datetime for the previous midnight in the time zone at the given datetime.
+* `timeComponent`: Gives the number of seconds since the previous midnight in the time zone at the given datetime.
+* `daysSinceEpoch`: Gives the number of Julian days since the beginning of the epoch at the given datetime.
+* `getOffset`: Gives the offset in seconds of the time zone from UTC at the given datetime, taking into account any daylight savings that may be being applied.
+* `getRawOffset`: Gives the base offset in seconds of the time zone, that is, without daylight savings applied.
+
+For usage information, see the API Reference for EPL (ApamaDoc).
```
kailol-sag, line 25: Still waiting for an answer on my question: "As the doc for PAM 1015.5 is not yet live, I cannot add a link. But is it possible for the user to see the updated ApamaDoc somewhere? Or should we remove the reference to the ApamaDoc?"

@skom-sag or @rthrippleton , do you know the answer ... ???

skom-sag: AFAIK, 10.15.5 ApamaDoc is not live. 

having said that this information on its own is not useful for the customer without knowing the API ... we don't have a published api doc or there's no way for a customer to inspect the API (ex: not available in designer for code completion / navigation etc). only an advanced user would be able to figure out with the capabilities offered by 'any' type.

I think the best option is merge this right away but remember to add a date/version number once PAM 10.15.5 is public. ex: May be you can add a note to the existing PAM 10.15.5 doc task for PAQ to also update this change log

kailol-sag: We said in our deployment meeting last week that we do not want to merge the change logs immediately. Instead we want to do this during the deployment meeting. 
If we merge this now, do we find this again when the PAM 10.15.5 doc is live? Do we have a good process for the change logs that are copied over from PAM? Trying to "remember" is not always the best thing to do ;-)
So if you still want to merge this NOW, then let me know.

rthrippleton: @kailol-sag - I agree, stick to the process, let's not make special one-off decisions. Only merge this based on a deployment that introduces the feature. And yes, remembering is bad!

Given that, I think the right thing to do is either
* Just forget about this entire changelog entry, and err on the side of not being misleading.
* Make it useful without needing an API reference link. So that means maybe we need a bit more detail in the changelog - almost like a precise of the API reference. Additionally, we _could_ say something like "This will be available in standalone Apama 10.15.5".

kailol-sag: Not sure what to do now. Please let me know what to add or remove. @Rob-J-SAG , what do you suggest?

Rob-J-SAG: I think go with Richard's second option. Add something like "This is available in Apama 10.15.5.", i.e. present tense and no "standalone".

kailol-sag: Thanks @Rob-J-SAG , I've added this. 
But I still don't know what to make of Richard's comment "So that means maybe we need a bit more detail in the changelog - almost like a precise of the API reference." ???

Rob-J-SAG: I don't think we need to add more detail about the API. What's currently written is enough to give devs a sense of what the new change brings.

rthrippleton: I'll clarify - somebody who is using EPL apps in streaming analytics doesn't have quite enough detail to use this feature _right now_, because the doc for Apama 10.15.5 isn't available. So this changelog entry needs to be a bit more verbose than the average changelog entry.

I'd suggest that @nmfa adds one example to this of just using one method, and also includes the method signatures (argument types/names and return type).

kailol-sag: Still waiting for a response from @nmfa

kailol-sag: Nick provided an example on https://itrac.eur.ad.sag/browse/PAM-34668
I copied that over into this change log entry.
All good now?

PR #1849, opened 2024-03-22
added change log entry re column compression for time series tenants

martin-sag:


pink-mongoose (commented):
As per discussion, we will:
1. not write a change log entry for this feature
2. include this information in the migration feature.

martin-sag:
As per comment, we close this PR. Announcement will happen later

PR #1901, opened 2024-04-03
PAQ change log for /prometheus improvement (PAB-4411)

kailol-sag:
This is for a fix. It won't show up on the doc website, so the exact release date does not matter.

Do we also need a "why" for this? If yes, then please provide input.


skom-sag:
I think redundant and performance already explains the why

content/change-logs/analytics/apama-in-c8y-20240319-Improved-implementation-of-prometheus-endpoint-in-Streaming-Analytics-microservices.md
```diff
@@ -0,0 +1,17 @@
+---
+date: 2024-03-19
+title: Improved implementation of the /prometheus endpoint in the Streaming Analytics microservices
```
Rob-J-SAG, line 3: "More efficient" instead of "Improved".
"Improved" is quite often simply too vague.

Rob-J-SAG (approved):

skom-sag:
eu-latest/zone-1 is updated to v25.93.0 today

PR #1953, opened 2024-04-17
BugherdTask_#817

dianaolives:
- Changed SmartREST 2.0 to SmartREST title because in the overview of the reference guide this section is referred to as such and it does not make sense of the section title to be SmartREST 2.0 and a subsection with that title as well. This is of course just a suggestion. I am still not sure because this whole section is about SmartREST 2.0.
- Overview subsection title changed to "Introduction" as well as file name changed to this.

I am not sure if it would be better to completely organize this section another way.

### DIFF
--- a/content/reference/smartrest-two-bundle/introduction.md
+++ b/content/reference/smartrest-two-bundle/introduction.md
@@ -1,6 +1,6 @@
 ---
 weight: 10
-title: Overview
+title: Introduction
 layout: redirect
 ---
 
--- a/content/reference/smartrest-two-bundle/smartrest-two.md
+++ b/content/reference/smartrest-two-bundle/smartrest-two.md
@@ -4,8 +4,6 @@ title: SmartREST 2.0
 layout: redirect
 ---
 
-### Overview
-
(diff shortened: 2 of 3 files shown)


BeateRixen:
> * Changed SmartREST 2.0 to SmartREST title because in the overview of the reference guide this section is referred to as such and it does not make sense of the section title to be SmartREST 2.0 and a subsection with that title as well. This is of course just a suggestion. I am still not sure because this whole section is about SmartREST 2.0.
> * Overview subsection title changed to "Introduction" as well as file name changed to this.
> 
> I am not sure if it would be better to completely organize this section another way.

I would not spend more time in reorganizing the section as this is "old" documentation. It is a bit better organized in the new documentation, further improvements should be done there. This documentation is owned by the Device & Control team, so  @l2p- should decide. 

@l2p- The issue here is that we have one heading "SmartREST 2.0" with a subheading "SmartREST 2.0" which confuses people.

image

BeateRixen:
In the new documentation we have a separate section for SmartREST, probably the structure is better here:
image

l2p- (approved):

BeateRixen (approved):

PR #1976, opened 2024-04-23
MTM-57921: Add section warning about notification 2.0 giving users th…

aborrett-SAG:
…e ability to subscribe to any device

### DIFF
--- a/content/reference/notifications-bundle/overview.md
+++ b/content/reference/notifications-bundle/overview.md
@@ -29,7 +29,7 @@ However, it does not yet support all of the notifications available from the Rea
 See the rest of this section and the detailed API documentation for full details of the notifications supported by this release of the Notifications 2.0 API.
 
 {{< c8y-admon-caution >}}
-Assigning a user Notification 2.0 access allows a user to create Notification 2.0 subscriptions to any device, including those which assigned inventory roles do not grant access to, bypassing inventory role RBAC.
+If you assign Notification 2.0 access to users, they can create Notification 2.0 subscriptions to any device, including those to which assigned inventory roles do not grant access, bypassing the inventory role RBAC.
 {{< /c8y-admon-caution >}}
 
 
--- a/content/reference/notifications-bundle/overview.md
+++ b/content/reference/notifications-bundle/overview.md
@@ -29,7 +29,7 @@ However, it does not yet support all of the notifications available from the Rea
 See the rest of this section and the detailed API documentation for full details of the notifications supported by this release of the Notifications 2.0 API.
 
 {{< c8y-admon-caution >}}
-If you assign Notification 2.0 access to users, they can create Notification 2.0 subscriptions and receive notification for any device, including those to which assigned inventory roles do not grant access, bypassing the inventory role RBAC.
+If you assign Notification 2.0 roles or permissions to users, they can create Notification 2.0 subscriptions and receive notifications for any device, including those to which assigned inventory roles do not grant access, bypassing the inventory role RBAC.
 {{< /c8y-admon-caution >}}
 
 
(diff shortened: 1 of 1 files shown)


content/reference/notifications-bundle/overview.md
```diff
@@ -28,6 +28,11 @@ New capabilities are added to Notifications 2.0 in each release of {{< product-c
 However, it does not yet support all of the notifications available from the Real-time notification API so it is not yet a complete replacement for the older API.
 See the rest of this section and the detailed API documentation for full details of the notifications supported by this release of the Notifications 2.0 API.
 
+{{< c8y-admon-caution >}}
+Assigning a user Notification 2.0 access allows a user to create Notification 2.0 subscriptions to any device, including those which assigned inventory roles do not grant access to, bypassing inventory role RBAC.
```
BeateRixen, line 32:
```suggestion
If you assign Notification 2.0 access to users, they can create Notification 2.0 subscriptions to any device, including those to which assigned inventory roles do not grant access, bypassing the inventory role RBAC.
```

content/reference/notifications-bundle/overview.md
```diff
@@ -28,6 +28,11 @@ New capabilities are added to Notifications 2.0 in each release of {{< product-c
 However, it does not yet support all of the notifications available from the Real-time notification API so it is not yet a complete replacement for the older API.
 See the rest of this section and the detailed API documentation for full details of the notifications supported by this release of the Notifications 2.0 API.
 
+{{< c8y-admon-caution >}}
+If you assign Notification 2.0 access to users, they can create Notification 2.0 subscriptions to any device, including those to which assigned inventory roles do not grant access, bypassing the inventory role RBAC.
```
sagscmi, line 32: I guess this should be changed to the same text you used in the equivalent OpenAPI PR?

slc-softwareag (approved):

content/reference/notifications-bundle/overview.md
```diff
@@ -226,7 +231,7 @@ The following summarizes the context and API support.
 ### Subscription filters
 Subscription filters provide fine-grained selection of the {{< product-c8y-iot >}} messages a subscription will forward.
 Filters can be provided as the **subscriptionFilter** field in a subscription's JSON object at creation time.
-It is a JSON object with **apis** and **typeFilter** fields.
```
gapa-c8y, line 229: I will have placed a double-space here deliberately (it won't have been a mistake) to break this up a bit. I don't think removing it is a problem however.

content/reference/notifications-bundle/overview.md
```diff
@@ -28,6 +28,11 @@ New capabilities are added to Notifications 2.0 in each release of {{< product-c
 However, it does not yet support all of the notifications available from the Real-time notification API so it is not yet a complete replacement for the older API.
 See the rest of this section and the detailed API documentation for full details of the notifications supported by this release of the Notifications 2.0 API.
 
+{{< c8y-admon-caution >}}
```
gapa-c8y, line 31: Thought: not sure about this, but we talk about these permissions just above (in the green requirements box). Should we be calling it ROLES or PERMISSIONS (maybe not in caps-case) as we do a few lines above, rather than referring to it as assigning access? Maybe we should say 'If you assign Notification 2.0 roles or permissions to users, ...' 

Also should '...receive notification ...' be pluralized to '...receive notifications ...'

gapa-c8y (approved):

BeateRixen (approved):

BeateRixen:
@aborrett-SAG Should this only go into the 10.18 documentation or also to newer versions? If so we need a second PR.

PR #2002, opened 2024-04-29
Change log for "On missing measurements create alarm" smart rule (PAB-4487)

kailol-sag:
@shrutidogra-23 , you wanted to define this as a "Fix", right?

### DIFF
--- a/content/change-logs/analytics/apama-in-c8y-20240429-Smart-rule-On-missing-measurements-create-alarm-now-sends-alarms-without-delay.md
+++ b/content/change-logs/analytics/apama-in-c8y-20240429-Smart-rule-On-missing-measurements-create-alarm-now-sends-alarms-without-delay.md
@@ -1,0 +1,17 @@
+---
+date:
+title: Smart rule "On missing measurements create alarm" now sends alarms without delay
+change_type:
+  - value: change-VSkj2iV9m
+    label: Fix
+product_area: Analytics
+component:
+  - value: component-M5-cepIIS
+    label: Streaming Analytics
+build_artifact:
+  - value: tc-KXXmo2SUR
+    label: apama-in-c8y
+ticket: PAB-4461
+version:
+---
+The "On missing measurements create alarm" smart rule now generates the alarms at the time period configured in the smart rule. Previously, the alarms were generated with a slight delay.
--- a/content/change-logs/analytics/apama-in-c8y-20240429-Smart-rule-On-missing-measurements-create-alarm-now-sends-alarms-without-delay.md
+++ b/content/change-logs/analytics/apama-in-c8y-20240429-Smart-rule-On-missing-measurements-create-alarm-now-sends-alarms-without-delay.md
@@ -14,4 +14,4 @@ build_artifact:
 ticket: PAB-4461
 version:
 ---
-The "On missing measurements create alarm" smart rule now generates the alarms at the time period configured in the smart rule. Previously, the alarms were generated with a slight delay.
+The "On missing measurements create alarm" smart rule now generates the alarms at the time period configured in the smart rule. Previously, the alarms were generated with a delay of approximately one minute.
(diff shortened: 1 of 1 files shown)


shrutidogra-23:
> @shrutidogra-23 , you wanted to define this as a "Fix", right?

content/change-logs/analytics/apama-in-c8y-20240429-Smart-rule-On-missing-measurements-create-alarm-now-sends-alarms-without-delay.md
```diff
@@ -0,0 +1,17 @@
...
+component:
+  - value: component-M5-cepIIS
+    label: Streaming Analytics
+build_artifact:
+  - value: tc-KXXmo2SUR
+    label: apama-in-c8y
+ticket: PAB-4461
+version:
+---
+The "On missing measurements create alarm" smart rule now generates the alarms at the time period configured in the smart rule. Previously, the alarms were generated with a slight delay.
```
Rob-J-SAG, line 17: I think we need more guidance than "slight" delay. What is the definition/range inferred here?

shrutidogra-23: There is a delay of a minute actually but I was avoiding quoting the number because in some weird edge case it might be more/less. But from what I saw while I ran the smart rule, alarms were delayed by a minute.

shrutidogra-23: Precisely in my scenario , I set the time period for 1 minute for the rule. Send a measurement at say 11.00, and then no further measurements were sent so was expecting alarms  to be raised at an interval of a minute but saw alarms at:
11:02
11:04
11:06 and so on

Rob-J-SAG: I think then the text can be changed to something like "Previously, the alarms were generated with approximately a minute's delay."

MWindel (approved):

content/change-logs/analytics/apama-in-c8y-20240429-Smart-rule-On-missing-measurements-create-alarm-now-sends-alarms-without-delay.md
```diff
@@ -0,0 +1,17 @@
...
+component:
+  - value: component-M5-cepIIS
+    label: Streaming Analytics
+build_artifact:
+  - value: tc-KXXmo2SUR
+    label: apama-in-c8y
+ticket: PAB-4461
+version:
+---
+The "On missing measurements create alarm" smart rule now generates the alarms at the time period configured in the smart rule. Previously, the alarms were generated with a delay of approximately one minute.
```
skom-sag, line 17: @shrutidogra-23, is the delay for a minute or the next polling window? i.e. if the rule is configured for 1 hr, is the alert still delayed by a minute or an hour ?

shrutidogra-23: No, it's not that bad :) 
we setup a listener which fires at every minute to check if the current time is over the configured time period. So the alarms were being delayed only by a minute (irrespective of what time period is set in rule config)

skom-sag: ok

skom-c8y (approved):

skom-c8y:
deployed to eu-latest on 8 Aug 2024

PR #2050, opened 2024-05-14
PAQ change log entry for infinite loop with pageSize (PAM-34761)

kailol-sag:


bph-sag (approved):

content/change-logs/analytics/apama-in-c8y-20240514-Queries-with-pageSize-set-to-1-no-longer-loop-infinitely.md
```diff
@@ -0,0 +1,17 @@
+---
+date:
+title: Queries with pageSize set to 1 no longer loop infinitely
```
Rob-J-SAG, line 3: Think we need to qualify queries as "REST queries" or "API queries". Apply the same to the description text too.

bph-sag: If there's any previous precedence for terminology here, go with that, I'd personally go with "REST queries".

kailol-sag: I change it to "REST queries".

Rob-J-SAG (approved):

skom-sag:
v25.167.0 is deployed to eu-latest on 11 June 2024

PR #2052, opened 2024-05-15
PAQ change log for transient network issues (PAB-4447)

kailol-sag:


content/change-logs/analytics/apama-in-c8y-20240515-Improved-robustness-of-Streaming-Analytics-against-transient-network-issues.md
```diff
@@ -0,0 +1,17 @@
...
+component:
+  - value: component-M5-cepIIS
+    label: Streaming Analytics
+build_artifact:
+  - value: tc-KXXmo2SUR
+    label: apama-in-c8y
+ticket: PAB-4447
+version:
+---
+The robustness of EPL apps and Analytics Builder models against transient network issues has been improved.
```
Rob-J-SAG, line 17: Second sentence.
"Specifically, if a temporary network issue causes a problem retrieving EPL apps from the inventory during recovery, then the EPL app is shown with an ERROR state in the UI, rather than the EPL app not shown at all."

gyan-sag: @Rob-J-SAG It won't be correct. Nothing will be shown if we cannot retrieve from the inventory in the first place - microservice will be restarted (existing behavior). Improvement is made around handling errors when trying to write to inventory. Most probably we should not mention what is shown in the UI because it depends on the when error occurs (PAB-4447 lists all cases). Failure to write to inventory during recovery means that EPL apps are visible in the UI without any errors (unless an EPL app itself had any issue found during deployment).

Rob-J-SAG: Please provide an alternative sentence, then. Karin and I can review. You need to give an example of what the fix has changed.

gyan-sag: I thought we agreed to be vague in yesterday's review meeting because we do not want to mention issues with platform/inventory and agreed on the original sentence.

Rob-J-SAG: You can't be _that_ vague. The description has to be identifiable to the fix. And the current description could apply to any network related robustness. Please provide an example of how the fix makes the EPL apps more robust. Don't need to go in to load balance issues, just keep with the terminology around network robustness.

gyan-sag: Sure. How about something following (apps are now more robust during recovery as well as normal saving, so removed the reference to the recovery):

"Specifically, if a temporary network issue causes a problem while saving EPL apps or Analytics Builder model to the inventory, then the EPP app or the Analytics Builder model is still displayed with correct state in the UI instead of it being displayed with incorrect state or not being displayed at all."

Rob-J-SAG: Like it - passes my test, over to Karin ;-)

Rob-J-SAG (approved):

skom-sag (approved):

skom-sag:
PAQ 25.137.0 is deployed to zone-1 / eu-latest today

PR #2251, opened 2024-07-05
Change my username in preparation for GHEC

scmi-c8y:


BeateRixen (approved):

BeateRixen:
@scmi-c8y I don´t know yet how this will be handled in the public repositories. Robert said that they will be taken care of after the private repos migration. I suggest to wait with this PR until we know more details.

scmi-c8y:
> @scmi-c8y I don´t know yet how this will be handled in the public repositories. Robert said that they will be taken care of after the private repos migration. I suggest to wait with this PR until we know more details.

@BeateRixen I already changed the username, so while I don't mind getting less mail from GitHub, this probably should be merged.

PR #2263, opened 2024-07-09
Proposal for limit page update

eickler:
This is a draft PR, general review is not yet required.

### DIFF
--- a/content/concepts/limits.md
+++ b/content/concepts/limits.md
@@ -22,7 +22,7 @@ The quotas listed here reflect the maximum values for the standard cloud subscri
 
 |Quota|Type|Value|
 |----|----|---:|
-|Tenant number|Soft|2000|
+|Tenant amount|Soft|2000|
 
 #### Microservices
 
@@ -46,6 +46,10 @@ The quotas listed here reflect the maximum values for the standard cloud subscri
 
 ### REST API
 
+|Quota|Type|Value|
+|----|----|---:|
+|Maximum API request duration|Hard|5 minutes|
+
 ### Realtime APIs
 
 |Quota|Type|Value|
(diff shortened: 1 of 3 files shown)


content/concepts/limits.md
```diff
@@ -1,28 +1,75 @@
...
-<tbody>
-<tr>
-<td style="text-align:left">Tenants</td>
-<td style="text-align:left">The maximum number of tenants supported is 2000.</td>
-</tr>
-</tbody>
-</table>
+Your {{< product-c8y-iot >}} service includes so-called ``quotas``. These quotas ensure that the service operates within the guarantees of the {{< product-c8y-iot >}} service-level agreements.
+
+There are three types of quotas in {{< product-c8y-iot >}}:
```
c8y-mayerc, line 11: the 3rd type of quotas is not present in the list below, is that intentional?

content/concepts/limits.md
```diff
@@ -1,28 +1,75 @@
...
+
+### Realtime APIs
+
+|Quota|Type|Value|
+|----|----|---:|
+|[MQTT message size](/device-integration/mqtt/)|Hard|16 KB|
+|[Notifications 2.0 message backlog](/change-logs/?#cumulocity-undefined-advance-notice-of-reduced-storage-limits-for-notifications-2.0-and-microservice-based-data-broker)|Hard|25 MB|
+|[Notifications 2.0 time-to-live](/change-logs/?#cumulocity-undefined-advance-notice-of-reduced-storage-limits-for-notifications-2.0-and-microservice-based-data-broker)|Hard|36 hours|
+|[Microservice-based data broker message backlog](/change-logs/?#cumulocity-undefined-advance-notice-of-reduced-storage-limits-for-notifications-2.0-and-microservice-based-data-broker)|Hard|50 MB|
+|[Microservice-based data broker time-to-live](/change-logs/?#cumulocity-undefined-advance-notice-of-reduced-storage-limits-for-notifications-2.0-and-microservice-based-data-broker)|Hard|36 hours|
```
scmi-c8y, line 61: These are technically all configurable and we document how to configure them, hence my question in the doc about whether "configurable" is a type of limit or a property of limits.

eickler: @scmi-c8y For me, there is a difference between what you can theoretically configure by manipulating internal configuration files and what we offer generally to be configurable. 

You can increase the size of message backlogs or the array limit in CDH, but I assume that the system has not been tested with the setup and there are consequences to system dimensioning (file system space? memory usage? -- otherwise the limit would not be there in the first place).

So rather than giving a carte blanche to change it, I would document it as hard limit. What we could do, though, is to offer a professional service to test for different hard limits for customers.

content/concepts/limits.md
eickler: Ah, well spotted. It was just the idea that we could have plans with different limits. I'll commit it out until we have that facility.

eickler:
I''ll close this pull request; further discussion can be carried out as part of PR #2452

PR #2268, opened 2024-07-10
[MTM-59297] Skip domain attribute in authorization Set-Cookie header

mbak-c8y:
While setting a new authorization cookie, the 'domain' attribute by default will be omitted. Skipped this attribute allows the browser to apply cookies only for current document URLs, not including subdomains. Behavior can be changed via a core property: 'auth.cookies.skip.domain: false'

### DIFF
--- a/content/change-logs/platform-services/cumulocity-10-20-472-0-skip-domain-attribute-in-authorization-set-cookie-header.md
+++ b/content/change-logs/platform-services/cumulocity-10-20-472-0-skip-domain-attribute-in-authorization-set-cookie-header.md
@@ -3,8 +3,8 @@ date: ""
 title: Skip domain attribute in authorization Set-Cookie header
 product_area: Platform services
 change_type:
-  - value: change-inv-3bw8e
-    label: Announcement
+  - value: change-VSkj2iV9m
+    label: Fix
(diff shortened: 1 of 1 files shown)


kgol-c8y (approved):

BeateRixen (approved):

BeateRixen:
@kgol-c8y I just changed your username in the codeowner file since you approved with the new username here :-).

mbak-c8y:
> it is unclear to me what this means for the average user. Can the user change a core property?
The average user shouldn't see any difference when logging in. Only in corner cases when someone uses a custom application that spreads cookies to sub-domains. Core property can only be changed by operation.

pink-mongoose:
So we need to provide information on what they do if they have an application that spreads cookies to sub-domains. We also need to make it clear that this is a security enhancement and remove the information about the core property as that is irrelevant to all SaaS customers, as I assume that core property is not at tenant level?

BeateRixen:
And we need an entry in the operations change logs, don´t we?

mbak-c8y:
Opened PR with rephrased RN https://github.com/SoftwareAG/c8y-docs/pull/2278/files

PR #2274, opened 2024-07-11
Fix hugo version update

carlosceia:
Hugo introduced some breaking changes in v0.123.0. I had to make some adjustments to fix that. The main changes regarding the content files are:

- The files located in the root of the content folder cannot match folder names, so these files names were changed to include the `-card` suffix.
- We had the `section` taxonomy, which conflicted with the built-in properties of Hugo, so it was renamed to `sector`
- Something similar happened to the `bundle` property we use in the front matter, which was changed to `bundlefolder`

Please check if everything works as expected on different platforms (linux, windows…) and with older versions of Hugo. 
I made the changes running the latest version on a mac: `hugo v0.128.2+extended darwin/amd64` which was installed with brew, and there's no easy way of switching versions :(


sag-eweingaertner (commented):
Checked out the branch locally. Works great for me.
Environment:
`$c8y-docs % hugo version
hugo v0.126.0+extended darwin/arm64 BuildDate=2024-05-14T13:24:11Z VendorInfo=brew
`

sag-eweingaertner (approved):

BeateRixen (approved):

BeateRixen:
Works for me as well. Using Chrome on a Mac and a very old Hugo version: Hugo Static Site Generator v0.76.2/extended darwin/amd64. Should perhaps update it :-).

kailol-sag (approved):
Works for me on Windows, with an older version of Hugo: Hugo Static Site Generator v0.80.0/extended windows/amd64

PR #2281, opened 2024-07-16
Update main.js

BeateRixen:


BeateRixen:
@carlosceia I modified the text in the banner but you please need to adjust the layout a bit as the height is too low when the banner text shows up in 2 rows on smaller screens.

PR #2359, opened 2024-08-13
MTM-57112 Add caution about reduced storage limits to user documentation

rohitjoshi-c8y:
This PR adds an important notice to the user documentation regarding upcoming changes to the storage limits for microservice-based data broker. I was told that customers rarely read the change logs, so perhaps this could help. Anyways, we have also contacted customers who are likely to be impacted by this change individually through CSMs, so they are covered.


content/data-broker/ms-data-broker-bundle/using-ms-data-broker.md
```diff
@@ -6,6 +6,17 @@ sector:
   - platform_administration
 ---
 
+{{< c8y-admon-important >}}
+##### Notice of reduced storage limits for Notifications 2.0 and microservice-based data broker
+
+**Effective Date:** July 1st, 2024
+
+Please be aware that the storage limits for microservice-based data broker will be significantly reduced. For detailed information about these changes and how they may impact your usage, please refer to [this announcement](https://cumulocity.com/docs/change-logs/?component=.component-messaging-service#cumulocity-undefined-advance-notice-of-reduced-storage-limits-for-notifications-2.0-and-microservice-based-data-broker).
```
BeateRixen, line 14: This should actually be exactly the other way round. The details should show up in the documentation and not in the change log (only) which will not be looked into after a while. Actually, details should always go into the documentation and the change log should only summarize the change. Since this was an announcement this might not have been possible, but now the details should definitely go into to the documentation. 

But in the documentation, please avoid phrasings like "has been reduced in July 2024". This will lead to outdated information in future releases. Just describe the storage limits. In the documentation of a specific release it is not really relevant that they have been reduced. This information has come with the announcement.

What is your take here @scmi-c8y ?

scmi-c8y: I agree that we should just cover the limits in the normal documentation and not refer back to the announcement.  We should also try to explain what can cause the limits to be hit and what (if anything) the customer can do to prevent this or resolve it when it does happen.

@rohitjoshi-c8y Will there also be a similar PR for Notifications 2.0?  IMHO it's probably more important to document the limits for that API given that it's the one where customers frequently get themselves into trouble.

content/data-broker/ms-data-broker-bundle/using-ms-data-broker.md
```diff
@@ -6,6 +6,21 @@ sector:
   - platform_administration
 ---
 
+{{< c8y-admon-important >}}
+##### Notice of current storage limits and message TTL for the microservice-based data broker
```
BeateRixen, line 10: Please move the heading outside of the important note

content/data-broker/ms-data-broker-bundle/using-ms-data-broker.md
```diff
@@ -6,6 +6,21 @@ sector:
...
+
+The **microservice-based data broker** stores messages persistently until they are successfully delivered to the destination tenant. To optimize resource usage, the storage limits for the microservice-based data broker have been adjusted as follows:
```
BeateRixen, line 12:
```suggestion
The microservice-based data broker stores messages persistently until they are successfully delivered to the destination tenant. To optimize resource usage, the storage limits for the microservice-based data broker have been adjusted as follows:
```

schmiel-cumulocity (approved):

BeateRixen:
@rohitjoshi-c8y @scmi-c8y What is the status of this PR?

rohitjoshi-c8y:
> @rohitjoshi-c8y @scmi-c8y What is the status of this PR?

@BeateRixen 
Sorry, lost track of this amidst everything else. I have moved the heading out, as you suggested. Thanks.

BeateRixen (approved):

gapa-c8y (approved):

tjegorow-c8y (approved):

BeateRixen:
@rohitjoshi-c8y @gapa-c8y Can this be merged finally?

rohitjoshi-c8y:
> @rohitjoshi-c8y @gapa-c8y Can this be merged finally?

Done.

PR #2374, opened 2024-08-21
PAB-4567 + PAB-4573: Streaming Analytics, breadcrumbs and changed names

kailol-sag:
This is for both https://itrac.eur.ad.sag/browse/PAB-4567 and https://itrac.eur.ad.sag/browse/PAB-4573.
Updated the doc/screenshots for the breadcrumbs that are now shown and the changed name on the "Model settings" and "App settings" links. 
Also changed the default names from "Model#" and "App#" to "New".


BeateRixen (approved):

skom-c8y (approved):
Deployed to eu-latest as part of apama-in-c8y v25.232.0 upgrade on 5th Sep 24

PR #2413, opened 2024-09-04
PAB-4603: Remove doc for alternative data source for model simulation

kailol-sag:
@skom-c8y , I'm assuming that this can be merged immediately.
Do we also want to auto-backport this to y2024?


BeateRixen (approved):

skom-c8y (approved):

skom-c8y:
Yes, we should merge this to develop right away.

and I would expect this to be back ported to all supported release trains so that customers on those older lines won't start using it.

kailol-sag:
Added the auto-backport labels for 2024. 

@skom-c8y  , do you also want to remove this from the 10.18 and 10.17 doc? If yes, then we have to add a separate PR because these versions are on the old "/guides/" website.

skom-c8y:
Yes, please do

PR #2422, opened 2024-09-06
PAB-4608: Sort icon in instance editor

kailol-sag:


BeateRixen:
@kailol-sag The image does not have the standard size (1440px for a full screen image).

BeateRixen (approved):

skom-c8y (approved):
This is deployed to eu-latest as part of apama-in-c8y v25.240.0

PR #2382, opened 2024-08-22
Fetch deployment status from build pipeline repo to update changelog dates

reey:
Just a proposal, execution details can be found here: https://github.com/reey/c8y-docs/actions/runs/10506995013

PR #2461, opened 2024-09-24
Mtm 59369/change logs automation

shaikhahme:
1. Added new workflow to trigger updating changelogs daily
2.  Updated changelogs.yml to better better log number of changed files
3. Removed redundant trigger workflows
4. Added different version formats handling
5. Replaced Environment variables and update repo link for c8y-iot-build-pipeline repository from forked to original
6. Replace manual token generation with actions/create-github-app-token@v1


eliasweingaertner:
@shaikhahme I really like this, as it looks simple and straightforward. Two questions:

a) How did you test it?
b) What happens if the python script runs into an error? Does the pipeline execution stop as expected?

shaikhahme:
> @shaikhahme I really like this, as it looks simple and straightforward. Two questions:
> 
> a) How did you test it? b) What happens if the python script runs into an error? Does the pipeline execution stop as expected?

a) I tested the scripts locally and also in GitHub actions
b) yes the pipeline execution stops, I also added error messages to make to help.

TLDR for test cases:
1. creates new PR for respective components when successful (https://github.com/shaikhahme/c8y-docs/pull/12)
2. workflow fails if error is raised in python script

you can see the workflow logs for different test scenarios from the comment in b821be05e1c89a8b28aff927cd3b279920ea8d9a

eliasweingaertner:
@shaikhahme asked me to mark this ready for review, as he has problems with Github.

Oezge-Akin (approved):

.github/workflows/daily-changelog-dates-trigger.yml
```diff
@@ -0,0 +1,82 @@
...
+          token=$(curl --fail --silent -X POST -H "Authorization: Bearer ${jwtToken}" -H "Accept: application/vnd.github.v3+json" "${githubApiUrl}/app/installations/${githubAppInstallationId}/access_tokens" | jq -r .token)
+          echo "jwt=${jwtToken}" 
+          echo "token=${token}" 
+          echo "jwt=${jwtToken}" >> "${GITHUB_OUTPUT}"
+          echo "token=${token}" >> "${GITHUB_OUTPUT}"
+
+      - name: Checkout c8y-iot-build-pipeline repository
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ vars.FORKED_REPO }}
```
reey: Why a `Forked` repo?

.github/workflows/update-changelog-dates.yml
reey, line 1: It seems a few of these changes are taken from: https://github.com/SoftwareAG/c8y-docs/pull/2382
Any reason why you've skipped some of these changes?

scripts/get-component-details-from-chart.py
reey, line 1: Any reason for rewriting https://github.com/SoftwareAG/c8y-docs/pull/2382/files#diff-4f3d2a35fb5ebe6c2282a25deabc2a777ccf68218838e1f164c47f8c97ae257fR64-R85 as a python script?
I at least think that we have more developers familiar with shell scripts compared to python to be honest..

add-release-dates/index.ts
```diff
@@ -32,19 +34,24 @@ async function processFile(filePath: string): Promise<boolean> {
...
-      "Version in file: ",
-      filePath,
-      "is not a valid semver. Skipping.."
-    );
+
+  if (!valid(data.version) && data.version.split('.').length==4){
+    const parts = data.version.split('.');
+    const normalizedVersion = `${parts[0]}${parts[1]}.${parts[2]}.${parts[3]}`;
+    data.version=normalizedVersion ;
+    }
```
reey: Can you log a warning here that we are trying to convert the none semver version into semver?

.github/workflows/daily-changelog-dates-trigger.yml
shaikhahme: @reey That is because I can't install an app on the https://github.com/Cumulocity-IoT/c8y-iot-build-pipeline repo. And I needed the app for authentication. 

or did I misunderstand how to use apps for authentication?

.github/workflows/update-changelog-dates.yml
shaikhahme: @reey No particular reason, I wanted to keep the changes to update-changelog-dates.yml minimal to keep testing easier, but I'll add the other changes too. thanks for catching it!

scripts/get-component-details-from-chart.py
shaikhahme: @reey this was a decision taken after talking to @eliasweingaertner , who suggested using python because we have more developers familiar with python than bash.

eliasweingaertner: Whoops. That was a misunderstanding: When @shaikhahme told me he was working on a bash script, I said that I'd prefer python. I didn't intend to have a working implementation replaced.

@reey Still, I believe the Python script is more readable. Are you okay with keeping that as well?

reey: I'm fine with a python script as long as someone maintains it (because I won't)..

.github/workflows/daily-changelog-dates-trigger.yml
reey: Okay, but in the future it won't be forked, right? So the name of the variable should not include `FORKED_` prefix..

scripts/get-component-details-from-chart.py
Oezge-Akin: @shaikhahme are you doing sth. additional than what is done in https://github.com/Cumulocity-IoT/c8y-docs/pull/2382/files#diff-4f3d2a35fb5ebe6c2282a25deabc2a777ccf68218838e1f164c47f8c97ae257fR64-R85

If it's same or easy to extend, then let's use the existing one to not duplicate things. 
If it is more complex then we can continue with python since it has good util libraries and it can make development/maintenance easier.

shaikhahme: Understood, since the python script has almost the same functionality as the bash script, I will use bash script.

.github/workflows/update-changelog-dates.yml
shaikhahme: resolved in a9b317e67f8967cfc8ee386f785248dd75dd51ce

.github/workflows/daily-changelog-dates-trigger.yml
shaikhahme: resolved in 1dc3707a3b05d181699b7313746dc6e46318e04e

scripts/get-component-details-from-chart.py
shaikhahme: resolved in 1dc3707a3b05d181699b7313746dc6e46318e04e

add-release-dates/index.ts
shaikhahme: resolved in 1dc3707a3b05d181699b7313746dc6e46318e04e

.github/workflows/daily-changelog-dates-trigger.yml
```diff
@@ -0,0 +1,76 @@
...
+          fetch-depth: 1
+
+      - name: Creating github access token
+        id: token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ vars.APP_ID }}
+          private-key: ${{ secrets.APP_PRIVATE_KEY }}
+          owner: cumulocity-iot
+          repositories: c8y-iot-build-pipeline
```
reey: Should allow us to avoid any changes in case of the next GitHub migration.
The `repositories` parameter is not required as the GitHub app we are using anyhow only has access to this single repo.
```suggestion
          owner: ${{ github.repository_owner }}
```

add-release-dates/index.ts
```diff
@@ -32,19 +34,25 @@ async function processFile(filePath: string): Promise<boolean> {
...
     return false;
   }
-  if (!valid(data.version)) {
-    console.debug(
-      "Version in file: ",
-      filePath,
-      "is not a valid semver. Skipping.."
-    );
+
+  if (!valid(data.version) && data.version.split('.').length==4){
```
reey: #trivial
```suggestion
  if (!valid(data.version) && data.version.split('.').length==4) {
```

add-release-dates/index.ts
```diff
@@ -32,19 +34,25 @@ async function processFile(filePath: string): Promise<boolean> {
...
+    const parts = data.version.split('.');
+    const normalizedVersion = `${parts[0]}${parts[1]}.${parts[2]}.${parts[3]}`;
+    console.warn("Non-semantic version: ", data.version ,"converted to semantic version format: ", normalizedVersion);
+    data.version=normalizedVersion ;
+    }
+
+  if(!valid(data.version)){
```
reey: #trivial
```suggestion
  }

  if(!valid(data.version)) {
```

add-release-dates/index.ts
```diff
@@ -32,19 +34,25 @@ async function processFile(filePath: string): Promise<boolean> {
...
+    console.debug("Version in file: ",filePath,"is not a valid semver. Skipping..");
     return false;
-  }
+    }
```
reey: #trivial
```suggestion
  }
```

scripts/get-component-details-from-chart.py
reey, line 1: It seems you've decided to use the shell script instead now? If so, this file is obsolete and can be removed, right?

shaikhahme: thanks for catching that. Resolved in ed86ba313b520018693f4edcfdefa896db1709ea

reey (approved):

PR #2503, opened 2024-10-11
Corrected clumsy section from REST (port 8443) with over MQTT (port 8…

logu-c8y:
…883) or REST (port 8443)


BeateRixen:
I will close this PR for now as we do not get any feedback and it is not that urgent.

PR #2512, opened 2024-10-14
Delete unused images

logu-c8y:


BeateRixen:
@logu1411 The text change in the file data-connectors should not show up here. You probably branched from your last branch? Please always branch from develop and remove this change from here.

PR #2547, opened 2024-10-22
added sector for each file where we want to see the options

logu-c8y:


BeateRixen:
@logu1411 This is a very good catch! Can you please check if this occurs in other sections of the documentation as well? I bet, for example, that we will have the same issue in the edge-kubernetes section.

logu-c8y:
> @logu1411 This is a very good catch! Can you please check if this occurs in other sections of the documentation as well? I bet, for example, that we will have the same issue in the edge-kubernetes section.

yeah sure. I will check it on other sections as well. May i know, should i have to create a new branch or i can commit my new changes into this branch?

BeateRixen:
> > @logu1411 This is a very good catch! Can you please check if this occurs in other sections of the documentation as well? I bet, for example, that we will have the same issue in the edge-kubernetes section.
> 
> yeah sure. I will check it on other sections as well. May i know, should i have to create a new branch or i can commit my new changes into this branch?

Please do all changes related to this bug in this branch here.

logu-c8y:
> > > @logu1411 This is a very good catch! Can you please check if this occurs in other sections of the documentation as well? I bet, for example, that we will have the same issue in the edge-kubernetes section.
> > 
> > 
> > yeah sure. I will check it on other sections as well. May i know, should i have to create a new branch or i can commit my new changes into this branch?
> 
> Please do all changes related to this bug in this branch here.

I have reviewed all the sections and found an issue only in the edge section. I have corrected it and made the necessary changes

BeateRixen (approved):

PR #2552, opened 2024-10-23
PAQ: Changed "Software AG Designer" to "Apama Plugin for Eclipse"

kailol-sag:
... and also renamed the shortcode file.

There are still other occurrences of "Software AG" such as the links to the Apama documentation. But this can only be changed when the Apama doc is available in Markdown and when I know the URL of the new doc website.

Similar changes (but we do not require a shortcode there) have been done on https://github.com/Cumulocity-IoT/c8y-docs-ops/pull/906 ,  which links to the #apama-plugin anchor defined in this PR.


BeateRixen (approved):

BeateRixen:
@kailol-sag @skom-c8y When is this change supposed to be merged?

themes/c8ydocs/layouts/shortcodes/apama-plugin.html
```diff
@@ -0,0 +1 @@
+{{- "Apama Plugin for Eclipse" -}}
```
skom-c8y: @kailol-sag  apama-plugin short code seems too generic and can be confused with other plugins like EPL plugin, Java plugin, connectivity plugin etc. can we instead use something like apama-plugin-for-eclipse ?

skom-c8y:
> @kailol-sag @skom-c8y When is this change supposed to be merged?

we are actively working on other doc topics, and we should be able to merge these in over the next few weeks.

themes/c8ydocs/layouts/shortcodes/apama-plugin.html
kailol-sag: @skom-c8y Ah, you're right about the confusion. I wanted to use something short. But I'll change this to apama-plugin-for-eclipse as suggested.

BeateRixen:
@skom-c8y This PR is only about the name change from Software AG Designer to Apama Plugin for Eclipse. I wonder if we can already merge this as we would like to reduce the number of SAG references throughout the documentation as much as possible until the end of the year. I know that there is more to do with all the links to the Apama documentation but we could implement these changes in a separate second step.

skom-c8y (approved):

skom-c8y:
> @skom-c8y This PR is only about the name change from Software AG Designer to Apama Plugin for Eclipse. I wonder if we can already merge this as we would like to reduce the number of SAG references throughout the documentation as much as possible until the end of the year. I know that there is more to do with all the links to the Apama documentation but we could implement these changes in a separate second step.

done

PR #2565, opened 2024-10-28
MTM-59369: Hotfixes for automation of changelog dates

shaikhahme:
1. Now checks if a PR to update dates is already exists, only creates a new PR if PR doesn't already exist.
2. Converts version to semantic for processing and back to non-semantic before writing to changelogs


add-release-dates/index.ts
```diff
@@ -56,6 +53,8 @@ async function processFile(filePath: string): Promise<boolean> {
     );
     return false;
   }
+
+  data.version = convertVersionFormat(data.version) // converts from semantic semantic to non-semantic version format
```
reey: now we would convert previously valid semver to a four digit version number.. We only want to have the initial version number back here.. if it was previously valid semver, it should stay like that..
So maybe just store the initial version number in a variable in line 37 and then afterwards just set `data.version` to the initial value..

shaikhahme: resolved in d1e7f030c7a8e43aa4cbb5855a9e7d11736e392a.

add-release-dates/index.ts
```diff
@@ -34,20 +34,17 @@ async function processFile(filePath: string): Promise<boolean> {
     console.warn("No version set in: ", filePath, "Skipping..");
     return false;
   }
+  originalVersion = data.version;
```
reey:
```suggestion
  const originalVersion = data.version;
```

add-release-dates/index.ts
```diff
@@ -56,6 +53,8 @@ async function processFile(filePath: string): Promise<boolean> {
     );
     return false;
   }
+
+  data.version = originalVersion
```
reey:
```suggestion
  data.version = originalVersion;
```

jakubsobolewskisag (approved):

reey (approved):

sourajitsaha17 (approved):

BeateRixen:
@shaikhahme @eliasweingaertner @reey Can this PR be merged so that we can proceed with the open PRs which add dates to various components? There are quite some change log entries that need to get published.

BeateRixen (approved):

BeateRixen:
> @shaikhahme @eliasweingaertner @reey Can this PR be merged so that we can proceed with the open PRs which add dates to various components? There are quite some change log entries that need to get published.

I can of course also add the dates manually for now if the automation is still work in progress.

PR #2647, opened 2024-11-15
Add Jane Porter to platform services change logs

kaczynskid-sag:
@JanePorter often rewrites the release notes to be presented to the customers, so by adding here to `CODEOWNERS` he can get notified and possibly improve the message before it's published.


BeateRixen:
@kaczynskid-sag JanePorter is already added.

BeateRixen:
@kaczynskid-sag Can we close this PR as Jane already was in the code owner file.
image

PR #2698, opened 2024-12-05
Digital Twin Manager application now uses title case

Vagdevi1998:
* Updated the application name for Digital Twin Manager to title case


BeateRixen:
@Vagdevi1998 There are more occurrences of "Digital twin manager" throughout the documentation. Please check and replace in all places, not only in the DTM documentation. Not sure how you want to handle older change log entries but probably you would want to keep the spelling there.

lara-c8y (approved):

BeateRixen (approved):

rahulc8y (approved):

PR #2819, opened 2025-01-16
Added script and workflows to check broken links

logu-c8y:


scripts/BrokenLinkChecker.js
pawel-rynarzewski-c8y, line 1: Would it be good to place it in the dedicated folder, instead of generic `scripts`? Like it's done for `add-release-dates` tool script.

scripts/BrokenLinkChecker.js
```diff
@@ -0,0 +1,117 @@
...
+  "product-c8y-iot": "Cumulocity IoT",
+  "c8y-edge-current-version-alt": "10.18",
+  "domain-c8y": "cumulocity.com",
+  "link-c8y-github": "https://github.com/Cumulocity-IoT",
+  "link-sag-portal": "https://empower.softwareag.com/",
+  "link-sag-tech-forum": "https://tech.forums.softwareag.com/",
+  "link-apamadoc-api": "https://documentation.softwareag.com/pam/10.15.5/en/webhelp/related/ApamaDoc/",
+  "link-sag-dev-community": "https://tech.forums.softwareag.com/tag/Cumulocity-IoT",
+  "link-apama-webhelp": "https://documentation.softwareag.com/pam/10.15.5/en/webhelp/pam-webhelp/",
+  "link-sag-privacy-statement": "https://www.softwareag.com/en_corporate/privacy.html",
```
pawel-rynarzewski-c8y: Should we consider `.softwareag.com` links as "broken", even if technically they are still working?

scripts/broken_links_markdown.csv
```diff
@@ -0,0 +1,19 @@
+URL,File Path,Status Code
+https://github.com/Cumulocity-IoT/cumulocity-examples/snmp,..\content\change-logs\device-management\SNMP-removed-from-docs.md,404
+https://download.cumulocity.com/,..\content\about_documentation\scope.md,Error
```
pawel-rynarzewski-c8y: I guess `https://download.cumulocity.com` links are fine, they just require login, which might caused error here

scripts/broken_links_markdown.csv
```diff
@@ -0,0 +1,19 @@
...
+https://download.cumulocity.com/,..\content\additional-resources\more-documentation.md,Error
+https://portal.azure.com/,..\content\datahub\integrating-datahub-with-other-products-bundle\integration-powerbi.md,403
```
pawel-rynarzewski-c8y: Requires login, but might be actually correct.

scripts/BrokenLinkChecker.js
BeateRixen: These are outdated and do no longer show up in the develop branch.

scripts/broken_links_markdown.csv
```diff
@@ -0,0 +1,19 @@
...
+https://github.com/eclipse/paho.mqtt.python/issues/354,..\content\device-integration\mqtt-examples-bundle\hello-mqtt-python.md,404
```
pawel-rynarzewski-c8y: It opens fine for browser user.

BeateRixen:
@logu1411 The links to download.cumulocity.com are not broken but require credentials.

scripts/broken_links_markdown.csv
```diff
@@ -0,0 +1,19 @@
...
+https://download.cumulocity.com/,..\content\edge\edge-databroker-bundle\databroker-edge.md,Error
+https://download.cumulocity.com/,..\content\edge\edge-databroker-bundle\databroker-edge.md,Error
+https://download.cumulocity.com/,..\content\edge\edge-update-bundle\updating-edge-gui.md,429
+https://download.cumulocity.com/,..\content\edge\edge-installation-bundle\prerequisites.md,429
+https://www.mathworks.com/help/predmaint/ug/remaining-useful-life-estimation-using-convolutional-neural-network.html,..\content\machine-learning\AI-ML-integration-bundle\step1-create-ai-ml-model.md,403
```
pawel-rynarzewski-c8y: works for browser user

scripts/broken_links_markdown.csv
```diff
@@ -0,0 +1,19 @@
...
+https://download.cumulocity.com/,..\content\glossary\y.md,429
+http://www.oracle.com/technetwork/java/javase/downloads/index.html,..\content\microservice-sdk\java-bundle\introduction.md,403
+https://www.oracle.com/technetwork/java/javase/downloads/index.html,..\content\microservice-sdk\java-bundle\ip-tracker-microservice.md,403
```
pawel-rynarzewski-c8y: They redirect but in the end reach download page.

scripts/broken_links_markdown.csv
```diff
@@ -0,0 +1,19 @@
...
+https://openmobilealliance.org/iot/lightweight-m2m-lwm2m,..\content\protocol-integration\lwm2m-bundle\introduction.md,404
+https://en.wikipedia.org/wiki/Sampling_(signal_processing,..\content\streaming-analytics\analytics-builder-bundle\details-of-values-and-blocks.md,404
+https://en.wikipedia.org/wiki/Quantization_(signal_processing,..\content\streaming-analytics\analytics-builder-bundle\details-of-values-and-blocks.md,404
+https://cumulocity.com/docs/<https://en.wikipedia.org/wiki/Glob_(programming,..\content\web\upgrade-bundle\c8y-cli.md,404
+https://code.visualstudio.com/,..\content\web\upgrade-bundle\updating-web-sdk-version.md,500
```
pawel-rynarzewski-c8y: It's not broken for browser user.

logu-c8y:
@pawel-rynarzewski-sag I have modified the short codes and moved the broken-link script and packages to a separate folder. Could you please check it and let me know, If i need to do anything?

.github/workflows/broken-links.yml
```diff
@@ -0,0 +1,68 @@
...
+                  })
+                  .join('\n')}`;
+              } else {
+                body = 'No broken links with 404 status found.';
+              }
+            } else {
+              body = 'Error: Broken link checker did not produce a CSV report.';
+            }
+
+            github.rest.issues.createComment({
```
pawel-rynarzewski-c8y: It creates a comment on every run. Usually we won't have broken links. But we will have a few pushes to the pull request which trigger `synchronize`. Maybe it would be better to only comment if there are some issues detected?

broken-links-script/broken_links_markdown.csv
pawel-rynarzewski-c8y, line 1: Is it intended to have this file committed? It looks like you're not going to update and commit it automatically, so it will soon be out-of-date. And it's only needed temporarily in the workflow, right?

broken-links-script/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "scripts",
```
pawel-rynarzewski-c8y: Update name?

broken-links-script/package.json
```diff
@@ -0,0 +1,19 @@
...
+  "version": "1.0.0",
+  "description": "- Node.js with version 20 and more is required to execute the scripts.",
```
pawel-rynarzewski-c8y: Is it accurate? In the workflow you set up Node.js 18.

broken-links.json
pawel-rynarzewski-c8y, line 1: What is this file for?

broken-links-script/package.json
```diff
@@ -0,0 +1,19 @@
...
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "node-fetch": "^3.3.2",
+    "p-limit": "^6.2.0"
+  },
+  "devDependencies": {
+    "linkinator": "^6.1.2"
```
pawel-rynarzewski-c8y: So in the end you're not using this dependency?

.github/workflows/broken-links.yml
```diff
@@ -0,0 +1,68 @@
...
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const fs = require('fs');
+            const path = './broken-links-script/broken_links_markdown.csv';
+            let body;
+            if (fs.existsSync(path)) {
+              const data = fs.readFileSync(path, 'utf8').trim();
+              const lines = data.split('\n').slice(1); // Skip the header line
+              const broken404Links = lines.filter(line => line.includes(',404')); // Filter only 404 errors
```
pawel-rynarzewski-c8y: This filtering is already done in `BrokenLinkChecker.js`, isn't it?

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,126 @@
...
+};
+
+const BASE_URL = "https://cumulocity.com/docs";
+
+const shortcodeMapping = {
+  "product-c8y-iot": "Cumulocity",
+  "c8y-edge-current-version-alt": "10.18",
+  "domain-c8y": "cumulocity.com",
+  "link-c8y-github": "https://github.com/Cumulocity-IoT/",
+  "c8y-edge-current-version-alt" : "10.18",
```
pawel-rynarzewski-c8y: `c8y-edge-current-version-alt` is duplicated

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,126 @@
...
+  "c8y-edge-current-version" : "1018",
+  "c8y-support-link" : "https://cumulocity.com/support",
+  "link-apamadoc-api" : "https://cumulocity.com/apama/docs/latest/related/ApamaDoc/index.html",
+  "link-c8y-doc-baseurl" : ".Page.Site.BaseURL",
+  "link-device-portal": "https://ecosystem.cumulocity.com/devices/?filter_cumulocity_certified=yes",
+  "c8y-support-link": "https://cumulocity.atlassian.net/servicedesk/customer/user/login?destination=portals",
```
pawel-rynarzewski-c8y: `c8y-support-link` is duplicated, and with different values

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,126 @@
...
+    }
+  } catch (error) {
+    return { url: fullUrl, file: mdFile, status: "Error" };
+  }
+
+  return null;
+};
+
+(async () => {
+  const projectDir = ".././content";
```
pawel-rynarzewski-c8y: `.` can be skipped here
```suggestion
  const projectDir = "../content"; 
```

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,126 @@
...
+  
+
+  const tasks = markdownFiles.map((mdFile) => {
+    const content = fs.readFileSync(mdFile, "utf8");
+    const links = [...content.matchAll(/(?<!\!)\[.*?\]\((.+?)\)/g)].map(
+      (match) => {
+        let link = match[1];
+        if (link.includes("(") && !link.endsWith(")")) {
+          link += ")"; // Add closing bracket if there's an opening bracket but no closing bracket
+        }
```
pawel-rynarzewski-c8y: You should rather not try to "repair" a link you are supposed to check, because you may miss the actual issue, see these examples, in both cases you would append `)` and verify them as valid, despite one not being correct in the source code.

broken-links-script/broken_links_markdown.csv
logu-c8y: yes, It is only needed temporarily. Once we work with the broken links, we need to correct it by checking this csv file locally.

broken-links.json
logu-c8y: I have checked few of the libraries to check the broken links. So I am deleting all the unnecessary files.

broken-links-script/BrokenLinkChecker.js
logu-c8y: yes, But for the Wikipedia pages, I have got the only solution. I have tried several regex expression to extract the correct link for Wikipedia pages.

broken-links.json
pawel-rynarzewski-c8y: So, this file is necessary or will be deleted?

broken-links-script/broken_links_markdown.csv
pawel-rynarzewski-c8y: If it's a temporary file, I'd not commit it. The broken links are listed in the created comment. And if really needed, the script can be run locally to generate the file.

broken-links-script/BrokenLinkChecker.js
pawel-rynarzewski-c8y: My point is that the current approach would miss a broken Wikipedia link, because it would add the missing parenthesis before validating it, but link in the source code will remain broken.

broken-links.json
logu-c8y: this file is not necessary as it seems like a binary file

broken-links-script/BrokenLinkChecker.js
logu-c8y: I researched this and found that by using encoded parentheses %28 and %29, we can define the Wikipedia URL correctly. Instead of using brackets, we can rely on these encodings without needing an extra closing parenthesis ). Is it okay?

broken-links-script/broken_links_markdown.csv
logu-c8y: Yes, we can run the script locally to check the broken links. Broken links csv file is not necessary to commit.

pawel-rynarzewski-c8y: So, can you remove it from PR?

broken-links.json
pawel-rynarzewski-c8y: Can you remove it from PR?

broken-links-script/BrokenLinkChecker.js
pawel-rynarzewski-c8y: Well, it would work for now, if you replace all occurrences, but you cannot guarantee that someone adding a new wikipedia link will remember to do the same. Have you checked the regex I proposed over chat (https://regex101.com/r/zo0LkU/1)? If so, have you noticed any issues with using it?

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,124 @@
...
+        return link;
+      }
+    );
+
+
+
+    return links.map((link) =>
+      limit(async () => {
+        const result = await checkLink(link, mdFile);
```
pawel-rynarzewski-c8y: It looks like you fire the check link via request separately for each occurrence of a link resulting in checking the same links multiple times, top ones:
```
https://cumulocity.com/docs/additional-resources/contacting-support/: 43
https://cumulocity.com/docs/standard-tenant/managing-permissions/: 18
https://cumulocity.com/docs/standard-tenant/ecosystem/#managing-applications: 16
https://cumulocity.com/docs/microservice-sdk/general-aspects/#microservice-manifest: 15
https://techcommunity.cumulocity.com/: 15
https://cumulocity.com/docs/device-integration/fragment-library/: 14
https://cumulocity.com/docs/device-management-application/monitoring-and-controlling-devices/#working-with-alarms: 14
https://cumulocity.com/docs/device-integration/device-integration-rest/: 13
https://cumulocity.com/docs/cockpit/working-with-dashboards/: 13
https://cumulocity.com/docs/streaming-analytics/analytics-builder/#editing-the-parameters-of-a-block: 13
https://cumulocity.com/docs/streaming-analytics/analytics-builder/#configuration: 10
https://cumulocity.com/docs/device-management-application/registering-devices/: 10
https://cumulocity.com/docs/standard-tenant/alarm-mapping/: 9
https://cumulocity.com/docs/cockpit/smart-rules/: 9
https://cumulocity.com/docs/get-familiar-with-the-ui/user-settings/: 8
https://cumulocity.com/api/core/#section/REST-implementation: 8
https://cumulocity.com/docs/enterprise-tenant/customization/#branding: 8
https://cumulocity.com/docs/device-management-application/managing-device-data/#managing-trusted-certificates: 8
https://cumulocity.com/api/core/#tag/Tenant-API: 8
https://cumulocity.com/docs/web/introduction/: 7
```
Maybe you could first gather the links from all files (keeping info on the occurrence location), and then run check for each link once?

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,124 @@
...
+
+  const resolvedLink = resolveHugoShortcode(link);
+  if (resolvedLink.startsWith("http://") || resolvedLink.startsWith("https://")) {
+    return resolvedLink;
+  }
+
+  return `${BASE_URL.replace(/\/$/, "")}/${resolvedLink.replace(/^\//, "")}`;
+};
+
+const checkLink = async (link, mdFile) => {
```
pawel-rynarzewski-c8y: Some links point to certain sections of a page, e.g.:
```
https://cumulocity.com/docs/microservice-sdk/general-aspects/#microservice-manifest: 15
https://cumulocity.com/docs/device-management-application/monitoring-and-controlling-devices/#working-with-alarms: 14
```
Now you only check if the whole page loads, right? Shouldn't we additionally verify that the target section actually exists on the page?

broken-links-script/BrokenLinkChecker.js
pawel-rynarzewski-c8y: @logu1411, looked into it again. Agreed that we should go with the encoding approach. So, let's encode the existing links with parentheses, and to prevent adding new ones in the future, let's make the script raise an error if it detects a link with unencoded parenthesis.

logu-c8y: yes, I will update the script

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,131 @@
...
+  const linkFileMap = new Map();
+  const uniqueLinks = new Set();
+
+  markdownFiles.forEach((mdFile) => {
+    const content = fs.readFileSync(mdFile, "utf8");
+    const links = [...content.matchAll(/(?<!\!)\[.*?\]\((.+?)\)/g)].map((match) => match[1]);
+
+    links.forEach((link) => {
+      if (hasUnencodedParentheses(link)) {
+        throw new Error(`Unencoded parentheses detected in link: "${link}" in file: ${mdFile}`);
```
pawel-rynarzewski-c8y: Wouldn't it be better not to throw an exception right away and stop any further validation, but rather collect any issues and list all of them in a comment (both about 404 and unencoded links)?

.github/workflows/broken-links.yml
```diff
@@ -0,0 +1,67 @@
...
+                  issue_number: context.issue.number,
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  body: body
+                });
+              } else {
+                console.log('No broken links with 404 status found');
+              }
+            } else {
+              console.log('Error: Broken link checker did not produce a CSV report.');
```
pawel-rynarzewski-c8y: What about exit codes? Shouldn't the workflow fail with non-zero code if any invalid links are found? Then it could be used e.g. as a pre-merge check.

broken-links.json
pawel-rynarzewski-c8y, line 1: If you do not use this file in your current approach, please do remove it from PR.

broken-links-script/broken_links_markdown.csv
pawel-rynarzewski-c8y, line 1: This file should not be committed into the source code. It's supposed to be generated every time from scratch when checking for broken links.

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,138 @@
...
+
+  const csvData =
+    "URL,Status Code,Files\n" +
+    brokenLinks.map((link) => `${link.url},${link.status},"${link.files.join("; ")}"`).join("\n");
+
+  fs.writeFileSync("broken_links_markdown.csv", csvData);
+  console.log("Broken links saved to broken_links_markdown.csv");
+
+  if (hasErrors) {
+    process.exit(1);
```
pawel-rynarzewski-c8y: I didn't mean to exit the script with non-zero code - this exits the workflow as well and no comment is even added to the pull request and developer has no access to the generated file either.

Rather we need to add the comment first and then fail the workflow to indicate that PR cannot be merged until the issues are resolved.

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,138 @@
...
+  if (brokenLinks.length > 0 || unencodedParenthesesIssues.length > 0) {
+    hasErrors = true;
+  }
+
+  const csvData =
+    "URL,Status Code,Files\n" +
+    brokenLinks.map((link) => `${link.url},${link.status},"${link.files.join("; ")}"`).join("\n");
+
+  fs.writeFileSync("broken_links_markdown.csv", csvData);
+  console.log("Broken links saved to broken_links_markdown.csv");
```
pawel-rynarzewski-c8y: Instead of generating `broken_links_markdown.csv` your script could generate `broken-links-warning.txt` whose contents could be directly forwarded to GitHub comment via yml file. That way you can include both 404 and unencoded parentheses issues in the same file and report them in the same comment.

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,135 @@
...
+  ));
+
+  let hasErrors = false;
+  
+  if (brokenLinks.length > 0 || unencodedParenthesesIssues.length > 0) {
+    hasErrors = true;
+  }
+
+  const txtData = 
+  "Broken Links Report:\n\n" +
```
pawel-rynarzewski-c8y: I'd drop this line, it unnecessarily becomes a point in the final comment.

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,135 @@
...
+
+  let hasErrors = false;
+  
+  if (brokenLinks.length > 0 || unencodedParenthesesIssues.length > 0) {
+    hasErrors = true;
+  }
+
+  const txtData = 
+  "Broken Links Report:\n\n" + 
+  brokenLinks.map(link =>
```
pawel-rynarzewski-c8y: The report is missing the unencoded parentheses issues.

You could consider unifying the way of gathering and reporting the issues, e.g.:

```
const issues = [];
...
issues.push({
  type: 'notFound404',
  url: link,
  files: linkFileMap.get(link)
});
...
issues.push({
  type: 'unencodedParentheses',
  url: link,
  files: linkFileMap.get(link)
});
```
and you could print them in structured way like:

### :warning: The following broken links were found:

| URL | Issue | Files |
| --- | --- | --- |
| https://github.com/Cumulocity-IoT//cumulocity-examples/snmp | notFound404 | <ul><li>../content/change-logs/device-management/SNMP-removed-from-docs.md</li><li>../content/change-logs/device-management/SNMP-removed-from-docs.md</li></ul> |
| https://en.wikipedia.org/wiki/Glob_(programming | unencodedParentheses | <ul><li>../content/web/upgrade-bundle/c8y-cli.md</li></ul> |

Then adding a new issue type would not require introducing a new list variable, but just using a new entry type.

You could write the whole report in markdown in `broken_links_report.md` (including both header and table/list), and just use it with no processing inside yml file:

```
if (fs.existsSync(path)) {
  const report = fs.readFileSync(path, 'utf8').trim();
  await github.rest.issues.createComment({
    issue_number: context.issue.number,
    owner: context.repo.owner,
    repo: context.repo.repo,
    body: report
  });
  core.setFailed('Broken links detected! You must resolve them before merging the PR.');
}
```

If no issues detected - just do not create a report file.
In case of unexpected exception while checking the links, the script will fail and it will fail the whole workflow too.

As for the file paths, if all start with `../content/`, maybe skip the beginning, because it only adds to the wall of text but does not bring any info.

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  return `${BASE_URL.replace(/\/$/, "")}/${resolvedLink.replace(/^\//, "")}`;
+};
+
+const hasUnencodedParentheses = (link) => {
+  return /[()]/.test(link);
+};
+
+const checkLink = async (link) => {
```
pawel-rynarzewski-c8y: I remember I raised a question about checking links with `#` section references. I see the thread is resolved, but I don't see any changes related to that, nor a reply on the thread. What's the status there?
https://github.com/Cumulocity-IoT/c8y-docs/pull/2819#discussion_r1936964358

.github/workflows/broken-links.yml
```diff
@@ -0,0 +1,57 @@
...
+      - name: Parse and Comment Broken Links
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const fs = require('fs');  
+            const reportPath = './broken-links-script/broken_links_report.md';
+
+            if (fs.existsSync(reportPath)) {
+              const report = fs.readFileSync(reportPath, 'utf8').trim();
+              const lines = report.split('\n\n');
```
pawel-rynarzewski-c8y: `lines` is obsolete

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,139 @@
...
+    const baseUrl = link.split("#")[0];
+    let response = await fetch(baseUrl, { method: "HEAD" });
+    if (response.status === 405) {
+      response = await fetch(baseUrl, { method: "GET" });
+    }
+    if (!response.ok && response.status !== 403 && response.status !== 500) {
+      return { type: "notFound404", url: link, status: response.status };
+    }
+  } catch (error) {
+    return { type: "error", url: link, status: "Network Error" };
```
pawel-rynarzewski-c8y: As you don't check what `error` is, it may indicate issues other than "Network Error". Maybe would be good to use `error.toString()` as a `type` of the issue, so that we have more info displayed in the summary table?
I see that `error` issue appeared for links like https://download.cumulocity.com or https://registry.c8y.io which are working correctly but require logging in. Could we distinguish such expected issues from other genuine errors?

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,139 @@
...
+  ));
+
+  if (issues.length > 0) {
+    let reportContent = "| URL | Issue | Files Location |\n";
+    reportContent += "| --- | --- | --- |\n";
+    reportContent += issues
+      .map((issue) =>
+        `| ${issue.url} | ${issue.type} | ${issue.files.join(", ")} |`
+      )
+      .join("\n");
```
pawel-rynarzewski-c8y: I admit that the table does not look good with long URLs and many file paths. Maybe we could keep the list but just align it a bit?
```suggestion
    const groupedIssues = issues.reduce((acc, issue) => {
      if (!acc[issue.type]) {
        acc[issue.type] = [];
      }
      acc[issue.type].push(issue);
      return acc;
    }, {});

    let reportContent = '### :warning: Broken links found!\n\n';
 
    for (const [type, issues] of Object.entries(groupedIssues)) {
      reportContent += `#### ${type}\n`;
      issues.forEach(issue => {
        reportContent += `- ${issue.url}\n`;
        issue.files.forEach(file => {
          reportContent += `  - ${file}\n`;
        });
      });
      reportContent += '\n';
    }

    return reportContent.trim();
```

Preview:

### :warning: Broken links found!

#### notFound404
- https://download.cumulocity.com
  - about_documentation/scope.md
  - additional-resources/more-documentation.md
  - edge/edge-databroker-bundle/databroker-edge.md
  - edge/edge-databroker-bundle/databroker-edge.md
  - edge/edge-installation-bundle/prerequisites.md
  - edge/edge-update-bundle/updating-edge-gui.md
  - glossary/y.md

#### unencodedParentheses
- https://en.wikipedia.org/wiki/Sampling_(signal_processing
  - streaming-analytics/analytics-builder-bundle/details-of-values-and-blocks.md

---

bonus points if file links could be clickable as well and lead to corresponding files on the current branch, where you could potentially edit the link right away:
```
const fileLink = `https://github.com/${repoOwner}/${repoName}/blob/${branchName}/${file}`;
reportContent += `  - [${file}](${fileLink})\n`;
```

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,139 @@
...
+        linkFileMap.get(resolvedLink).push(relativePath);
+      }
+      if (hasUnencodedParentheses(link)) {
+        issues.push({
+          type: "unencodedParentheses",
+          url: link,
+          status: "Invalid Character",
+          files: [relativePath],
+        });
+      }
```
pawel-rynarzewski-c8y: This is actually another check of the link, right? You could have it at the beginning of `checkLink`, therefore avoiding reporting urls like `https://en.wikipedia.org/wiki/Quantization_(signal_processing` twice in the summary.

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,139 @@
...
+    const links = [...content.matchAll(/(?<!\!)\[.*?\]\((.+?)\)/g)].map((match) => match[1]);
+
+    links.forEach((link) => {
+      const resolvedLink = resolveFullUrl(link);
+      if (resolvedLink) {
+        uniqueLinks.add(resolvedLink);
+        if (!linkFileMap.has(resolvedLink)) {
+          linkFileMap.set(resolvedLink, []);
+        }
+        linkFileMap.get(resolvedLink).push(relativePath);
```
pawel-rynarzewski-c8y: I see for some URLs file paths are duplicated. Maybe check if a file is already in the list before adding it again?

broken-links-script/BrokenLinkChecker.js
```diff
@@ -0,0 +1,139 @@
...
+  return `${BASE_URL.replace(/\/$/, "")}/${resolvedLink.replace(/^\//, "")}`;
+};
+
+const hasUnencodedParentheses = (link) => {
+  return /[()]/.test(link);
+};
+
+const checkLink = async (link) => {
+  try {
+    const baseUrl = link.split("#")[0];
```
pawel-rynarzewski-c8y: This will just get rid of `#section` part from the URL, right?

Note that we can have at least 2 types of internal links:

```
- **Alarms** - Displays alarms for the device, see also [Working with alarms](/device-management-application/monitoring-and-controlling-devices/#working-with-alarms).
```
This one will be correctly resolved to https://cumulocity.com/docs/device-management-application/monitoring-and-controlling-devices/#working-with-alarms. However, by executing `fetch` you will only confirm that https://cumulocity.com/docs/device-management-application/monitoring-and-controlling-devices can be loaded, but not if the anchor with id `working-with-alarms` actually occurs on the target page, so you won't catch if the `id` was renamed or removed.

```
see [To enable/disable a smart rule](#to-enabledisable-a-smart-rule) for details
```
This one will be incorrectly resolved to https://cumulocity.com/docs/#to-enabledisable-a-smart-rule which misses the middle part. `fetch` will only confirm that the main page https://cumulocity.com/docs can be loaded, but it tells nothing about the actual target section.

broken-links-script/BrokenLinkChecker.js
logu-c8y: When i run the script, Now i am getting this only link (https://download.cumulocity.com/) as an error. when i open it  on the browser it is redirecting to (https://login.cumulocity.com/u/login?) For this type of links what should i do?

pawel-rynarzewski-c8y: Sorry, missed this comment. I see `fetch` is not able to follow the redirections correctly. Is it still an issue with the cypress approach, or is cypress able to follow such a link correctly and determine that login page was displayed?

PR #2827, opened 2025-01-20
DTM 2025.0 changelog

lara-c8y:
In this PR, we have entries of changes after 2024.0. It excludes entries for the fixes provided in CD for new features introduced after 2024.0.


lara-c8y:
@BeateRixen , should we do this change directly here in 2025 branch or first in develop and then graft the change to here?

BeateRixen:
@lara-c8y This is a good point. I would want to leave it up to you if you only remove the tickets from the 2025 release branch or from develop as well. Removing them from develop as well would mean to remove them entirely from the repo which is ok for me, as you might also come up with a different approach in the future and not add them at all to the repo.

BeateRixen (approved):

spau6 (approved):

PR #2882, opened 2025-02-04
Fix change log date format inconsistencies for apama-in-c8y [no-issue]

skom-c8y:


yta-c8y (approved):

BeateRixen:
@yta-c8y @skom-c8y Both date formats actually work. The longer one is automatically created if you use Netlify to add the entries.

BeateRixen (approved):

PR #2955, opened 2025-02-18
Improvement of streaming analytics introduction section

eickler:
When i read the introduction to streaming analytics with a "fresh mind", I noticed some oddities/unclarities. 
 * It should start with an explanation on what streaming analytics is rather than with what Cumulocity does and what it's domain model is. That is described elsewhere in sufficient detail.
* The paragraph "These user-defined operations…"  uses the word "operations" in three meanings: User-defined logic, REST API calls and actual operations in the Cumulocity domain model sense of commands sent to devices. 
* In the paragraph "The Cumulocity platform includes an Apama correlator…", none of the terms introduced there is explained.

So I wanted to propose these small improvements.

### DIFF
--- a/content/streaming-analytics/introduction-analytics.md
+++ b/content/streaming-analytics/introduction-analytics.md
@@ -7,11 +7,8 @@ weight: 10
 aliases:
   - /streaming-analytics/overview-analytics/
 ---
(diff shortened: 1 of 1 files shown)


skom-c8y (approved):
Looks sensible

BeateRixen (approved):

PR #3021, opened 2025-03-05
Edge installation changes for minor issues found in user testing

rthrippleton:
- In some environments, the newly installed k3s wasn't on the path, make the path explicit.
- We found that in some cases, it's really important to follow the k3s OS requirements e.g. the RHEL default firewall settings upsets k3s.
- If you apply an Edge CR immediately after installing the operator, there is a race condition resulting in an error where the webhook can't be reached. Changing the `helm upgrade` command to use `--wait` means that the operator installation waits until all services (including the webhook) are ready.

### DIFF
--- a/static/files/edge-k8s/c8yedge-operator-install.sh
+++ b/static/files/edge-k8s/c8yedge-operator-install.sh
@@ -136,4 +136,5 @@ helm upgrade --install c8yedge-operator oci://${C8YEDGE_REGISTRY_HOST}/edge/helm
     --set image.pullPolicy="${C8YEDGE_IMAGE_PULL_POLICY}" \
     --set imageCredentials.registry="${C8YEDGE_REGISTRY_HOST}" \
     --set imageCredentials.username="${C8YEDGE_REGISTRY_USERNAME}" \
-    --set imageCredentials.password="${C8YEDGE_REGISTRY_PASSWORD}"
+    --set imageCredentials.password="${C8YEDGE_REGISTRY_PASSWORD}" \
+    --wait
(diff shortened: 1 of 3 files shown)


BeateRixen (approved):

static/files/edge-k8s/c8yedge-operator-install.sh
```diff
@@ -136,4 +136,5 @@ helm upgrade --install c8yedge-operator oci://${C8YEDGE_REGISTRY_HOST}/edge/helm
     --set image.pullPolicy="${C8YEDGE_IMAGE_PULL_POLICY}" \
     --set imageCredentials.registry="${C8YEDGE_REGISTRY_HOST}" \
     --set imageCredentials.username="${C8YEDGE_REGISTRY_USERNAME}" \
-    --set imageCredentials.password="${C8YEDGE_REGISTRY_PASSWORD}"
+    --set imageCredentials.password="${C8YEDGE_REGISTRY_PASSWORD}" \
+    --wait
```
bbyreddy, line 140: - [ ] This is a good addition. We should also add this option to the corresponding command in the run.py, Makefile where we do the edge operator install or upgrade.

bbyreddy (approved):

static/files/edge-k8s/c8yedge-operator-install.sh
rthrippleton: Already in the PR for the PySys-changes

BeateRixen:
@rthrippleton Can this PR be merged?

rthrippleton:
> @rthrippleton Can this PR be merged?

It can now, as the user has confirmed that one of the key changes here actually worked for them :-).

PR #3046, opened 2025-03-07
add broken link script

logu-c8y:


broken-links-script/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "broken-links-script",
+  "version": "1.0.0",
+  "main": "BrokenLinkChecker.js",
```
pawel-rynarzewski-c8y: This file no longer exists.

broken-links-script/cypress/fixtures/example.json
pawel-rynarzewski-c8y, line 1: I'd remove this example file.

broken-links-script/cypress/screenshots/link-checker.cy.js/Link and Routing Validation - Batching URLs -- should verify all URLs in batches and capture errors without stopping (failed).png
pawel-rynarzewski-c8y, line 1: I'd not commit these test results into repo, but we could definitely make these screenshots available for download from workflow run artifacts.

broken-links-script/all_links.json
pawel-rynarzewski-c8y, line 1: Do not commit a file which is supposed to be regenerated from scratch before every use.

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -0,0 +1,175 @@
...
+    if (unencodedParenthesisPattern.test(url)) {
+      if (!brokenLinks.some(link => link.url === url)) {
+        brokenLinks.push({ url, files, status: "Adjust unencoded parentheses" });
+      }
+      return true;
+    }
+    return false;
+  };
+
+  it('should verify all URLs in batches and capture errors without stopping', function () {
```
pawel-rynarzewski-c8y: If you have just a single `it` block which executes checks for all links, if there is a failure on any of them, you'll only get a single screenshot from the very last step which probably won't show how the problematic target page looked like. You'd rather need to create `it` block for each link and check them individually. This may allow you to get a nice cypress test report that can be linked from the comment, like we do for `qa-verify` label in `cumulocity-ui` repo. btw, I see there is no github workflow in this PR, is it intended?
What is the purpose of having batches here? Is their execution parallelized or otherwise optimized?

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -0,0 +1,175 @@
...
+      fragment,
+      fragment.replace(/\//g, '-'),
+      fragment.replace(/\//g, '_')
+    ];
+    return variants.some((variant) => {
+      const regexId = new RegExp(`id=["']${variant}["']`);
+      const regexHref = new RegExp(`href=["']#${variant}["']`);
+      return regexId.test(htmlContent) || regexHref.test(htmlContent);
+    });
+  };
```
pawel-rynarzewski-c8y: Why not use cypress' `get`/`find`?

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: Once the script is finalized, I'll create a new pull request to add a GitHub workflow file that identifies any broken lists in the pull request

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: I used RegExp in checkFragmentExists to match fragment variants (e.g / to - or _) in id or href within raw HTML (doc.body.innerHTML). get() or find() would be less flexible for this string-based check.

pawel-rynarzewski-c8y: Not sure I got it, if we have a hash with `/`, like `Section/header`, then we should expect an element with exactly this id, because otherwise the browser won't scroll to the desired element. Or am I missing the point here?

broken-links-script/package.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "broken-links-script",
+  "version": "1.0.0",
+  "type": "module",
+  "scripts": {
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "test:headless": "cypress run --browser chrome --spec 'cypress/e2e/link-checker.cy.js'"
```
pawel-rynarzewski-c8y: Why not use the main `test` script to run your tests, and maybe also include the prerequisite command (we want the tests to always run with the latest state of links, right?)?
```suggestion
    "test": "node Extractlinks.js && cypress run --browser chrome --spec 'cypress/e2e/link-checker.cy.js'"
```

broken-links-script/Extractlinks.js
```diff
@@ -0,0 +1,130 @@
...
+    headerAnchorMatches.forEach(match => {
+      const anchor = "#" + match[1];
+      const resolvedAnchorLink = resolveFullUrl(anchor, relativePath, content);
+      if (resolvedAnchorLink) {
+        if (!linkMap[resolvedAnchorLink]) {
+          linkMap[resolvedAnchorLink] = new Set();
+        }
+        linkMap[resolvedAnchorLink].add(relativePath);
+      }
+    });
```
pawel-rynarzewski-c8y: Can you explain this part? Are you finding every header across all md files, even if it's not a link itself, nor is it being linked to? That adds a lot of additional url checks for headers that are never linked to from our own docs... right?

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -0,0 +1,188 @@
...
+      let baseUrl = url;
+      let fragment = null;
+      let isCodexPage = false;
+      let isFragmentCheckRequired = false;
+
+      if (validateUrl(item)) {
+        expect(validateUrl(item)).to.be.false;
+        completedTests++;
+        return;
+      }
```
pawel-rynarzewski-c8y: `validateUrl(item)` is already checked in the `if` as truthy, so this assertion will always fail. I guess your intention was to fail the test when the given URL is not valid?

As you switched to cypress for running the tests, maybe we could make better use of it. I'd urge you to rely more on the features of Cypress and do not duplicate its work by tracking errors on your own in counters and arrays, just let it do its job. See this example:

```js
  const expectValidUrl = (url) => {
    const unencodedParenthesisPattern = /[()]/;
    if (unencodedParenthesisPattern.test(url)) {
      throw Error(`Unencoded parentheses in URL: ${url}`);
    }
  };

  Cypress.on('fail', (error) => {
    // quickly came up with this way to pass the source files to the error handler, do not really like it, maybe there is a better way?
    const sourceFiles = Cypress.env('sourceFiles');
  
    if (sourceFiles) {
      const sourceFilesList = `This URL was used in the following files:\n- ${sourceFiles.join('\n -')}`;
      error.message += `\n\n${sourceFilesList}`;
    }

    throw error;
  });

  urls.forEach((item) => {
    it(`should validate URL: ${item.link}`, () => {
      Cypress.env('sourceFiles', item.files);

      cy.log(`Starting test for URL: ${item.link}`);
      let url = item.link;
      let baseUrl = url;
      let fragment = null;
      let isCodexPage = false;
      let isFragmentCheckRequired = false;

      expectValidUrl(url);
```

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -0,0 +1,188 @@
...
+        cy.get('[data-cy="c8y-title--title-outlet"] .text-truncate', { timeout: 10000 }).then(($el) => {
+          cy.document().then((doc) => {
+            const targetElement = $el[0];
+            if (!targetElement) {
+              brokenLinks.push({
+                url,
+                files: item.files, 
+                status: 'Target view missing [data-cy="c8y-title--title-outlet"] .text-truncate'
+              });
+              expect(targetElement, 'Target element should exist').to.exist;
```
pawel-rynarzewski-c8y: Then you'd not need to track such errors on your own, the cypress test will fail if it does not find desired element within given timeout and will output a relevant error.

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -0,0 +1,188 @@
...
+      cy.log(`Finished test for URL: ${item.link}`);
+    });
+  });
+
+  afterEach(() => {
+    cy.log(`Current progress: ${completedTests}/${totalTests}`);
+  });
+
+  after(() => {
+    cy.log(`Final: completedTests: ${completedTests}, totalTests: ${totalTests}`);
```
pawel-rynarzewski-c8y: No need to generate report on your own, cypress will do that for you, just make sure that error messages from cypress are meaningful.

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -0,0 +1,188 @@
...
+                url,
+                files: item.files, 
+                status: 'Target view missing [data-cy="c8y-title--title-outlet"] .text-truncate'
+              });
+              expect(targetElement, 'Target element should exist').to.exist;
+            } else if (isFragmentCheckRequired && fragment) {
+              if (!checkFragmentExists(doc.body.innerHTML, fragment)) {
+                fragmentErrors.push({ url, files: item.files, fragment });
+                expect(checkFragmentExists(doc.body.innerHTML, fragment), 'Fragment should exist in the page').to.be.true;
+              }
```
pawel-rynarzewski-c8y: Here you could do the check in a helper function, like `expectFragmentExists(fragment)`.
It'd be similar to `checkFragmentExists` but would directly throw an error if fragment is not found.

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -0,0 +1,188 @@
...
+
+  const checkFragmentExists = (htmlContent, fragment) => {
+    const escRegExp = (string) => string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+    const escFragment = escRegExp(fragment);
+  
+    const regexId = new RegExp(`id=["']${escFragment}["']`);
+    if (fragment.includes('/')) {
+      return regexId.test(htmlContent);
+    }
+    const regexHref = new RegExp(`href=["']#${escFragment}["']`);
```
pawel-rynarzewski-c8y: Why do you check for a link containing given fragment? Wouldn't it give us false positive result if a link is on the page, but the corresponding header which is supposed to be on the same page too is not present?

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -0,0 +1,188 @@
+const urls = require('../../all_links.json');
+describe('Link and Routing Validation - Individual URL Checks', () => {
+  const brokenLinks = [];
+  const fragmentErrors = [];
+  let completedTests = 0;
+  let totalTests = 0;
+
+  const checkFragmentExists = (htmlContent, fragment) => {
+    const escRegExp = (string) => string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
```
pawel-rynarzewski-c8y: Do you really need innerHTML and regexp here, it looks like you now search by exact id, so a check like ``` cy.get(`#${fragment}`).should('exist') ``` should suffice, right?

broken-links-script/Extractlinks.js
logu-c8y: There are few links that we are using with header. For example, in c8y-docs-ops: the link we are referring based on the header such as http://localhost:1313/docs-ops/operation-load-balancer/load-balancer-configuration/#keep-alive-timeout.

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: If we use the get method to check the fragments, the links which are having fragments are getting failed

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: For some of the links the href is required for example: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#upgrade-to-spring-boot-3 here using "href" they are point to the section.

broken-links-script/Extractlinks.js
pawel-rynarzewski-c8y: sure, that's an example from another repo, but we have such cases in c8y-docs as well, e.g.:

https://github.com/Cumulocity-IoT/c8y-docs/blob/92489a52ac14f347e6871ff854020473416f9c20/content/device-integration/sigfox-bundle/managing-connectivity-settings.md?plain=1#L70-L76

However, my point is these are not active links but just anchors. See what's happening here: you find an anchor in source code, you generate a link to it, and later you'll verify if this generated link leads to the anchor, but note that this generated link may never actually occur in the c8y-docs, so it means you verify non-existing link (see the example above, this header is never linked to from other md files in c8y-docs).

And if this header happens to be linked to from another repo, you cannot verify if their link is correct, they'd need their own test to check it.

Therefore, I'd focus on identifying active links in c8y-docs and verifying them.

broken-links-script/cypress/e2e/link-checker.cy.js
pawel-rynarzewski-c8y: I see, you're right, in this case they indeed add custom prefix to ids and it's harder to verify... but the point about false positives is still valid. Let's take this page for example: https://cumulocity.com/docs/cockpit/smart-rules-collection/#smart-rule-variables

It contains a link:
```html
<div class="content">In certain rule parameters, various trigger fields can be used as variables, see <a href="#smart-rule-variables">Smart rule variables</a>.</div>
```
which is supposed to lead to this header:
```html
<h3 id="smart-rule-variables">Smart rule variables<button class="btn-link bookmark" data-clipboard-text="#smart-rule-variables">
  <span class="fa dlt-c8y-icon-link"></span>
</button> </h3>
```
But if we now remove the header or change its id, without updating the link's `href`, your script will count it as valid, because it sees the `href`, although the link is actually broken.

Could we take care of both cases?

broken-links-script/Extractlinks.js
logu-c8y: In the future, there's a possibility that this type might be used. Since the same script also needs to be integrated into the c8y-docs-ops repository, using such a feature later on could lead to some links being missed, right?

broken-links-script/cypress/e2e/link-checker.cy.js
pawel-rynarzewski-c8y: ah, I see, because in these cases you make a `cy.request` to get the page's HTML, instead of actually loading it with `cy.visit`. Honestly, I'd unify the approach to testing various types of links as much as possible: always load them with `cy.visit` and allow it to fail on error status codes (it's the default, so you may just skip `failOnStatusCode`), then depending on the target page you may check additional assertions, e.g. you can test fragments differently depending on how the target page uses them. And again, do not track the error yourself, just rely on cypress error reporting.

Then this block of code:
```ts
      if (url.includes('/api/')) {
        if (url.includes('#')) {
          const [baseUrl, fragment] = url.split('#');
          cy.visit(url, { failOnStatusCode: false, timeout: 10000 }).then(() => {
            cy.get('body').then(($body) => {
              const expectedHref = `#${fragment}`;
              if ($body.find(`a[href="${expectedHref}"]`).length === 0) {
                brokenLinks.push({
                  url,
                  files: item.files, 
                  status: `Expected anchor with href "${expectedHref}" not found`
                });
                expect($body.find(`a[href="${expectedHref}"]`).length, 'Anchor with expected href should exist').to.be.greaterThan(0);
              }
              completedTests++;
            });
          });
        }
```

could be written just as an `if` with single assertion:

```ts
  const expectNoUnencodedParentheses = (url) => {
    cy.wrap(url).should('not.match', /[()]/, `URL should not contain unencoded parentheses: ${url}`);
  };

  urls.forEach((item) => {
    it(`should validate URL: ${item.link}`, () => {
      Cypress.env('sourceFiles', item.files);
      const url = item.link;
      const fragment = url.split('#')[1];

      expectNoUnencodedParentheses(url);

      cy.visit(url, { timeout: 10000 });

      if (url.includes('/api/') && fragment) {
        cy.get(`[id="${fragment}"]`).should('exist', `Element with ID "${fragment}" should exist`);
      }
```

If you follow this for other tested target pages, you will end up with much simpler code which makes use of cypress features to report issues instead of having to track them yourself.

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: While checking the ID, I tested it by changing the ID, and it worked. For some other links, I changed the header, and those also showed up correctly in the broken links list. So in both cases, it worked as expected.

broken-links-script/Extractlinks.js
pawel-rynarzewski-c8y: Sorry, not sure I got your point. Did you mean that although currently we may not have active links to some of these headers, we might add such links later on? Of course, it's possible, but these new links will be verified by the part where you extract the links, not the anchors. If you meant something else, best to have a short call to clarify it before your last working day.

BeateRixen: @logu-c8y If you refer to the anchors inside of text (as opposed to the heading IDs) , didn´t we say that we will simply not use them?

broken-links-script/cypress/e2e/link-checker.cy.js
pawel-rynarzewski-c8y: This is not what I meant.
Let me try to demonstrate the potential issue I have in mind with this side-PR: https://github.com/Cumulocity-IoT/c8y-docs/pull/3269
In case of further questions, let's catch on a call.

broken-links-script/Extractlinks.js
```diff
@@ -0,0 +1,130 @@
...
+};
+
+const BASE_URL = "https://cumulocity.com/docs";
+
+const shortcodeMapping = {
+  "product-c8y-iot": "Cumulocity",
+  "domain-c8y": "cumulocity.com",
+  "link-c8y-github": "https://github.com/Cumulocity-IoT/",
+  "c8y-edge-current-version-alt": "10.18",
+  "c8y-edge-current-version": "1018",
```
pawel-rynarzewski-c8y: The current values of these shortcodes are defined somewhere else, this is not their source of truth, right? Wondering what happens if the version is updated for the deployed docs, but the script still extracts and tests the older one. Any way to avoid it? Or it won't be an issue?

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -0,0 +1,104 @@
...
+      
+        // Check that the page loaded and isn't a 404
+        cy.get('[data-cy="c8y-title--title-outlet"] .text-truncate', { timeout: 20000 })
+          .invoke('text')
+          .should('not.be.empty')
+      
+        // Validate the URL matches exactly what was provided (including fragment)
+        cy.url().should('eq', url);
+      
+;
```
pawel-rynarzewski-c8y: remove `;`?

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -0,0 +1,104 @@
...
+          expect(response.status).to.equal(200);
+          expectFragmentExists(response.body, fragment);
+        });
+      }
+
+      else {
+        cy.request({ url, timeout: 10000 }).then((response) => {
+          expect(response.status).to.equal(200);
+        });
+      }
```
pawel-rynarzewski-c8y: I think that:
```ts
cy.visit(url, { timeout: 20000 });

if (fragment) {
  cy.get(`[id="${fragment}"]`, { timeout: 10000 }).should('exist', `Element with ID "${fragment}" should exist`);
}
```
should be the default way of checking a link (because it's closer to the way how a user will interact with an url in their browser), only fall back to direct REST requests if there is a real need.

logu-c8y: That makes a lot of sense. I agree that using cy.visit provides a more accurate representation of how a user would interact with the page, especially for checking fragment IDs

broken-links-script/Extractlinks.js
logu-c8y: As of now, whenever the shortcode mappings change (e.g., due to version updates in the deployed docs), we need to manually update them in the script as well. I considered this the most straightforward approach for now, especially since the shortcode usage is relatively stable.

.github/workflows/link-checker.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Install Dependencies
+        run: |
+          cd broken-links-script
+          npm install
+
+      - name: Generate links
+        run: |
+          cd broken-links-script
+          node Extractlinks.js
```
pawel-rynarzewski-c8y: Not needed, already included in `npm test`.

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -0,0 +1,130 @@
...
+      
+        cy.document().then((doc) => {
+          const escRegExp = (string) => string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+          const escFragment = escRegExp(fragment);
+          const regexId = new RegExp(`id=["']${escFragment}["']`);
+      
+          const html = doc.documentElement.innerHTML;
+          const found = regexId.test(html);
+      
+          expect(found, `Fragment "${fragment}" should exist in HTML as an id`).to.be.true;
```
pawel-rynarzewski-c8y: Duplicated code from helper function?

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -0,0 +1,137 @@
...
+  });
+
+  urls.forEach((item) => {
+    it(`should validate URL: ${item.link}`, () => {
+      const url = item.link;
+      const fragment = url.includes('#') ? url.split('#').slice(-1)[0] : null;
+      const isCodexPage = url.includes('codex/#/');
+      const isApiPage = url.includes('/api/');
+      const isGithubPage = url.includes('github.com');
+      const nonHtmlExtensions = ['.txt','.json','.pdf','.zip','.csv','.xml','.not','.bin','.dat','.tar','.gz','.rar','.xsd','.yaml'];
```
pawel-rynarzewski-c8y: I think we also have at least 1 link to a `.pot` file? Should it be listed here too?

pawel-rynarzewski-c8y (approved):

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -60,7 +58,7 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
       const isCodexPage = url.includes('codex/#/');
       const isApiPage = url.includes('/api/');
       const isGithubPage = url.includes('github.com');
-      const nonHtmlExtensions = ['.txt','.json','.pdf','.zip','.csv','.xml','.not','.bin','.dat','.tar','.gz','.rar','.xsd','.yaml'];
+      const nonHtmlExtensions = ['.txt','.json','.pdf','.zip','.csv','.xml','.not','.bin','.dat','.tar','.gz','.rar','.xsd','.yaml','pot'];
```
pawel-rynarzewski-c8y: I would add a dot `.pot` for consistency with the other extensions here and to avoid accidental matching with a non-file url that just happens to end with `pot`

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -6,15 +6,12 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
...
-    const regexId = new RegExp(`id=["']${escFragment}["']`);
-    return regexId.test(htmlContent);
-  };
 
   const expectFragmentExists = (htmlContent, fragment) => {
-    expect(checkFragmentExists(htmlContent, fragment), `Fragment "${fragment}" should exist in HTML`).to.be.true;
+    const escFragment = escRegExp(fragment);
+    const regex = new RegExp(`(id=["']${escFragment}["']|name=["']${escFragment}["'])`);
+    const exists = regex.test(htmlContent);
+    expect(exists, `Fragment "${fragment}" should exist in HTML`).to.be.true;
```
pawel-rynarzewski-c8y: Maybe better to explicitly write what was looked for? e.g.: `An element with id or name "${fragment}" should exist in HTML`, otherwise it might be unclear for unfamiliar developers what we mean by a "fragment".

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -39,9 +36,10 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
     cy.document().then((doc) => {
       const html = doc.documentElement.innerHTML;
       expectFragmentExists(html, fragment);
-      
       const ids = Array.from(doc.querySelectorAll('[id]')).map(el => el.id);
-      cy.log(`Available IDs:\n${ids.join('\n')}`);
+      const names = Array.from(doc.querySelectorAll('a[name]')).map(a => a.getAttribute('name'));
+      const allFragments = [...ids, ...names];
+      cy.log(`Available fragments on page:\n${allFragments.join('\n')}`);
```
pawel-rynarzewski-c8y: Similar here: `Available elements on page with ids and names:`

BeateRixen (approved):

PR #3173, opened 2025-04-03
MTM-62975 - Adding more information of measurements fragments

dhbivanova:
Adding definition for fragments and series, fixing the measurements creation example

### DIFF
--- a/content/concepts/domain-model-bundle/measurements.md
+++ b/content/concepts/domain-model-bundle/measurements.md
@@ -30,7 +30,6 @@ Measurements consist of a time when the measurement was taken, the unique identi
 Similar to the inventory model, fragments are used to identify characteristics of particular devices. Each measurement fragment is an object that holds the actual measurements as properties, also known as series. The property name corresponds to the name of the measurement and includes two properties:
 
 -   value: The individual measurement value that is required for every series.
-
 -   unit: The unit associated with the measurement.
 
 In the example above, a three-phase electricity meter sends readings for the different electrical phases. A measurement fragment maps the names of the individual readings (for example, "A+" or "A-") to the actual numeric value and unit of the measurement.
--- a/content/concepts/domain-model-bundle/measurements.md
+++ b/content/concepts/domain-model-bundle/measurements.md
@@ -33,7 +33,7 @@ Similar to the inventory model, fragments are used to identify characteristics o
 
 -   unit: The unit associated with the measurement.
 
-In the above example, a three-phase electricity meter sends readings for the different electrical phases. In the example above, such a fragment maps the names of the individual readings (for example, "A+" or "A-") to the actual numeric value and unit of the measurement.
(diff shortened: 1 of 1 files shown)


mbrz-c8y (approved):

eickler (commented):
Looks good, one question: The last sentence reads "Readings can hold various additional information that applications may require."

(Maybe @martin-c8y ) What is our strategy with respect to "additional information in measurements"?  Can we have clear guidance here what users should put here? As far as I understood, we would like the user to create measurements that are compatible with our timeseries model and deprecate the very liberal model going forward. 

So if there is "additional information" that is compatible with the timeseries model, I would put it here; if the additional information is not compatible with the timeseries model, I would remove the comment. 

The same applies also to the REST API docs and the cumulocity-clients-java. The Java client encourages to use also
    private ValueType type;
    private String quantity;
    private StateType state;
If that is not compatible with our desired timeseries model, we should mark these as \@Deprecated.

content/concepts/domain-model-bundle/measurements.md
```diff
@@ -11,22 +11,29 @@ Measurements consist of a time when the measurement was taken, the unique identi
...
 ```
 
-Similar to the inventory model, fragments are used to identify characteristics of particular devices. In the above example, a three-phase electricity meter sends readings for the different electrical phases. In the example above, such a fragment maps the names of the individual readings (for example, "A+" or "A-") to the actual numeric value and unit of the measurement.
+Similar to the inventory model, fragments are used to identify characteristics of particular devices. Each measurement fragment is an object that holds the actual measurements as properties, also known as series. The property name corresponds to the name of the measurement and includes two properties:
+
+-   value: The individual measurement value that is required for every series.
+
+-   unit: The unit associated with the measurement.
+
+In the above example, a three-phase electricity meter sends readings for the different electrical phases. In the example above, such a fragment maps the names of the individual readings (for example, "A+" or "A-") to the actual numeric value and unit of the measurement.
```
BeateRixen:
```suggestion
In the example above, a three-phase electricity meter sends readings for the different electrical phases. A measurement fragment maps the names of the individual readings (for example, "A+" or "A-") to the actual numeric value and unit of the measurement.
```

content/concepts/domain-model-bundle/measurements.md
```diff
@@ -11,22 +11,29 @@ Measurements consist of a time when the measurement was taken, the unique identi
...
+        "A-:1": { "value": 2, "unit": "kWh" }
     }
 }
 ```
 
-Similar to the inventory model, fragments are used to identify characteristics of particular devices. In the above example, a three-phase electricity meter sends readings for the different electrical phases. In the example above, such a fragment maps the names of the individual readings (for example, "A+" or "A-") to the actual numeric value and unit of the measurement.
+Similar to the inventory model, fragments are used to identify characteristics of particular devices. Each measurement fragment is an object that holds the actual measurements as properties, also known as series. The property name corresponds to the name of the measurement and includes two properties:
+
+-   value: The individual measurement value that is required for every series.
+
```
BeateRixen:
```suggestion
```

BeateRixen (approved):

martin-c8y (approved):

PR #3207, opened 2025-04-10
[MTM-62399] based on feedback this change is important enough that it…

mbak-c8y:
Based on feedback and discussion within the team, this change is important enough to be published as an 'Announcement' instead of just a fix. I suspect that posting this as an announcement might help it reach more people.
@BeateRixen @JanePorter, is that okay with you?


BeateRixen (approved):

BeateRixen:
> Based on feedback and discussion within the team, this change is important enough to be published as an 'Announcement' instead of just a fix. I suspect that posting this as an announcement might help it reach more people. @BeateRixen @JanePorter, is that okay with you?

Yes and approved.

kowa-c8y (approved):

JanePorter (approved):

PR #3253, opened 2025-04-23
remove HDFS occurrences

logu-c8y:


rthrippleton (commented):
What's the background for this? Is CDH removing support for HDFS in a later fix on y2025?

PR #3269, opened 2025-04-28
[test only] Show issue with href and broken links

pawel-rynarzewski-c8y:


pawel-rynarzewski-c8y:
@logu-c8y, this PR demonstrates an issue I had in mind when writing https://github.com/Cumulocity-IoT/c8y-docs/pull/3046#discussion_r2038534372:
- I changed an anchor id of one of the headers: `#smart-rule-variables` -> `#smart-rule-variables-changed`
- however, the script still validates the existing but not updated link to this anchor (`cockpit/smart-rules-collection/#smart-rule-variables`) as valid:
Peek 2025-04-28 16-34

For testing purposes:
- I changed the base docs url to use the actual preview of this PR, instead of docs currently deployed in production
- added a filter to only check links with `smart-rule` to avoid running through all links

logu-c8y (commented):
I have updated the check to only look for actual IDs. Now, broken links that use href="#..." won't be marked as valid anymore.

pawel-rynarzewski-c8y:
> I have updated the check to only look for actual IDs. Now, broken links that use href="#..." won't be marked as valid anymore.

ok, except for `github.com` where we indeed must check for `href` instead.

PR #3276, opened 2025-05-02
MTM-63162 adding the charachters limitation for the latest value tenant option

dhbivanova:


content/standard-tenant/managing-data-bundle/latest-value.md
```diff
@@ -20,7 +20,9 @@ PUT /tenant/options/measurement.series.latestvalue
   // or "*":"" to enable all
 }
 ```
-where the key is a filter of measurement series that must be persistent and its value must always be an empty string (left for a future use case).
+where the key is a filter of measurement series that must be persistent and its value must always be an empty string (left for a future use case).  
+
+**⚠️ Important:** Property names used for fragment and series must not contain whitespaces nor the special characters ``` $ & + , / : ; = ? @ " < > # % { } | \ ^ ~ [ ] ` ```. This is necessary to ensure the new tenant option is processed correctly and saved successfully.
```
BeateRixen:
```suggestion
{{< c8y-admon-important >}}
Property names used for fragment and series must not contain whitespaces nor the special characters ``` $ & + , / : ; = ? @ " < > # % { } | \ ^ ~ [ ] ` ```. This is necessary to ensure the new tenant option is processed correctly and saved successfully.
{{< /c8y-admon-important >}}
```

martin-c8y (approved):

BeateRixen (approved):

PR #3332, opened 2025-05-16
Added script and github workflow for checking broken links in 2025 re…

logu-c8y:


.github/workflows/link-checker.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 'latest'
+
+      - name: Install Dependencies
+        run: |
+          cd broken-links-script
+          npm install
```
pawel-rynarzewski-c8y: Use `npm ci`.

.github/workflows/link-checker.yml
```diff
@@ -0,0 +1,33 @@
...
+
+      - name: Run Cypress tests
+        run: |
+          cd broken-links-script
+          npm test
```
pawel-rynarzewski-c8y: It needs a running hugo server, doesn't it? You probably need similar approach as in https://github.com/Cumulocity-IoT/c8y-docs-ops/pull/1067/files.

.github/workflows/link-checker.yml
logu-c8y: changed to npm ci

.github/workflows/link-checker.yml
logu-c8y: In c8y-docs develop branch, we have the same approach which is running properly.

pawel-rynarzewski-c8y: As discussed in the call, we don't need to run hugo here because we check directly against published version of the docs.

pawel-rynarzewski-c8y (approved):

BeateRixen (approved):

BeateRixen (approved):

PR #3380, opened 2025-05-27
MTM-63619 latest values announcement and new toggle description

dhbivanova:


BeateRixen:
@dhbivanova @martin-c8y If I get it right this feature has now changed from preview to GA. Shouldn´t we phrase it accordingly in the change log entry? The version selected, should then refer to the version when the feature went to GA and not when it was originally introduced. At least this is my take. What do you think, @JanePorter ?

content/standard-tenant/managing-data-bundle/latest-value.md
```diff
@@ -93,6 +93,33 @@ GET /inventory/managedObjects?withLatestValues=true&query=$filter=c8y_LatestMeas
   ]
 }
 ```
+
+However, in scenarios where measurements are delayed in arriving (due to network latency or other factors), the system may incorrectly display them as the latest measurement, even though they are technically out of order.
```
JanePorter: Remove 'However', it is not needed.

JanePorter (approved):
Good information, just one requested alteration that does not change the actual information.

content/standard-tenant/managing-data-bundle/latest-value.md
dhbivanova: It's gone :)

dhbivanova:
> @dhbivanova @martin-c8y If I get it right this feature has now changed from preview to GA. Shouldn´t we phrase it accordingly in the change log entry? The version selected, should then refer to the version when the feature went to GA and not when it was originally introduced. At least this is my take. What do you think, @JanePorter ?

@BeateRixen I've changed the release version to match the version in this ticket: https://cumulocity.atlassian.net/browse/MTM-51189 . Can you please review the PR again?
TA

pkepa-c8y (approved):

content/change-logs/platform-services/cumulocity-2025-120-0-Latest-measurement-values-can-be-stored-as-part-of-a-device-managed-object.md
```diff
@@ -0,0 +1,18 @@
...
+product_area: Platform services
+component:
+  - value: component-JlFdtOPva
+    label: REST API
+build_artifact:
+  - value: tc-QHwMfWtBk7
+    label: cumulocity
+version: 1017.128.0
+---
+Support of automated persistence of measurement values under the `c8y_LatestMeasurements` fragment has been introduced. If a measurement is created with a series that matches the configuration the device managed object is updated with the last series sent to the platform.
```
BeateRixen:
```suggestion
Support of automated persistence of measurement values under the `c8y_LatestMeasurements` fragment, which has previously been introduced as a preview feature, is now generally available. If a measurement is created with a series that matches the configuration the device managed object is updated with the last series sent to the platform.
```

BeateRixen: @dhbivanova I discussed this with @neuerburg and we agreed that we should mention that this feature transitions from preview to GA. Also, please use the version number with which the feature flag has been modified as this is the version number relevant for this transition change here.

content/change-logs/platform-services/cumulocity-2025-120-0-Latest-measurement-values-can-be-stored-as-part-of-a-device-managed-object.md
```diff
@@ -0,0 +1,18 @@
+---
+date: 2025-05-27
```
BeateRixen:
```suggestion
date:
```

BeateRixen: Will be added automatically

content/change-logs/platform-services/cumulocity-2025-120-0-Latest-measurement-values-can-be-stored-as-part-of-a-device-managed-object.md
BeateRixen: @neuerburg We have the special situation here that there is no software change (any feature flag) involved. It was just a decision that this feature now transitions from preview to GA. It is (and always was) configurable via tenant options. Should we still use the version from which this mode changed? I´d say so.

dhbivanova:
@BeateRixen as discussed I've updated the version to a recent one. Can you please review again?

BeateRixen (approved):

PR #3452, opened 2025-06-13
Fix broken link

logu-c8y:


BeateRixen:
@logu-c8y I wonder why this has not been detected earlier as this is not a new text?

BeateRixen (approved):

logu-c8y:
> @logu-c8y I wonder why this has not been detected earlier as this is not a new text?

I remember this link was flagged as broken earlier. I must have missed fixing it when going through the previous batch of broken links.

PR #3457, opened 2025-06-17
no-issue change the wording to avoid the 'late arrival' phrase

dhbivanova:
Change the wording to avoid the 'late arrival' phrase

### DIFF
--- a/content/standard-tenant/managing-data-bundle/latest-value.md
+++ b/content/standard-tenant/managing-data-bundle/latest-value.md
@@ -95,7 +95,7 @@ GET /inventory/managedObjects?withLatestValues=true&query=$filter=c8y_LatestMeas
 ```
 
 In scenarios where measurements are delayed in arriving (due to network latency or other factors), the system may incorrectly display them as the latest measurement, even though they are technically out of order.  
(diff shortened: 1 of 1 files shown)


martin-c8y (approved):

content/standard-tenant/managing-data-bundle/latest-value.md
```diff
@@ -95,7 +95,7 @@ GET /inventory/managedObjects?withLatestValues=true&query=$filter=c8y_LatestMeas
 ```
 
 In scenarios where measurements are delayed in arriving (due to network latency or other factors), the system may incorrectly display them as the latest measurement, even though they are technically out of order.  
-To address this, we've introduced a new toggle, `strongConsistency`. When this toggle is enabled (set to true), late-arriving measurements will not be shown as the latest data for the device, regardless of when they were actually received. Instead, only measurements that arrive in the correct order will be treated as the latest, ensuring that the most accurate, timely data is always presented.  
+To address this, we've introduced a new toggle, `strongConsistency`. When this toggle is enabled (set to true), the out-of-order measurements will not be shown as the latest data for the device, regardless of when they were actually received. Instead, only measurements that arrive in the correct order will be treated as the latest, ensuring that the most accurate, timely data is always presented.
```
BeateRixen:
```suggestion
To address this, the toggle `strongConsistency` is provided. When this toggle is enabled (set to true), the out-of-order measurements will not be shown as the latest data for the device, regardless of when they were actually received. Instead, only measurements that arrive in the correct order will be treated as the latest, ensuring that the most accurate, timely data is always presented.  
```

BeateRixen (approved):

PR #3467, opened 2025-06-18
rephrase the context

logu-c8y:
This is a BugHerd Ticket
Error Description: This text is a bit misleading. It implies that devices will "magically" be configured with this information but instead this is the information that needs to be configured on the device based on the information from the csv file.

### DIFF
--- a/content/device-management-application/registering-devices-bundle/bulk-device-registration.md
+++ b/content/device-management-application/registering-devices-bundle/bulk-device-registration.md
@@ -54,7 +54,7 @@ ID;CREDENTIALS;TYPE;NAME;ICCID;IDTYPE;PATH;SHELL;AUTH_TYPE
 006064ce8077;OowoGKAbiNJs;c8y_Device;Sample_Device2;+491555555;c8y_Serial;bulk group/subgroup2;1;BASIC
 ```
 
-To connect the devices, they are pre-registered with the relevant information. More specific, each device will be configured as follows:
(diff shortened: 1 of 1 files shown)


hame-c8y (approved):

BeateRixen (approved):

BeateRixen:
@l2p- Please approve. 
@logu-c8y Please do not merge before Philipp has approved.

l2p- (commented):
this change makes no difference. The section is just as misleading as before because the device needs to be pre-provisioned with the credentials or certificate and not anything else from the csv. We should highlight that this process process is for registering devices with pre-existing credentials as opposed to other types of registrations where we provision the device with credentials.

logu-c8y:
> this change makes no difference. The section is just as misleading as before because the device needs to be pre-provisioned with the credentials or certificate and not anything else from the csv. We should highlight that this process process is for registering devices with pre-existing credentials as opposed to other types of registrations where we provision the device with credentials.

Could you please suggest which text would be suitable here to make this clearer?

BeateRixen:
@l2p- Should we finally merge this as is or will you provide any suggestion how to make it better?

l2p-:
I think this should be closed without merging.

both the initial request and this change seem to stem from an incorrect understanding of the mechanism. For bulk registration the device "magically" comes with credentials already and we just tell the platform what they are for the connection to happen. There is no configuration of anything happening during that registration process.

PR #3501, opened 2025-06-30
no issue: editing the announcement for the previos values

dhbivanova:


martin-c8y (approved):

JanePorter (changes requested):
The 'c8y_Previous Measurements' fragment is reserved for internal system use only and may not be used by external users.

JanePorter (approved):

PR #3577, opened 2025-07-28
Added change log parameters for dm

BeateRixen:
- Added missing build artifacts and components for the change logs related to Device Management according to this list: https://docs.google.com/spreadsheets/d/1n9RUxJUc0otC3Tq7qlfNVrvxppFXjoqEkt5KhN2ObBI/edit?pli=1&gid=0#gid=0
- Resorted the entries alphabetically.

Due to the resorting, the added entries are hard to detect.

- Added components:
- Advanced Software Management, Device simulator, Grid Migration
- Added build artifacts:
- grid-migration, sms-gateway, sigfox-agent

I left out device-parameter for now (in accordance with Gopi) as it is not yet relevant. Will add it as soon as it is needed to avoid too many empty categories.


l2p- (approved):

data/component.json
```diff
@@ -81,17 +91,18 @@
...
-    },{
-      "id": "component-wWIzHJ12j",
-      "option": "Cloud Fieldbus",
+    },
+    {
+      "id": "component--LJtTuzaN",
+      "option": "Device Parameter",
       "product_area": [
         {
           "value": "product_area-m1iHjqikD",
```
gosh89, line 108: Missing team lable here?

BeateRixen: This is not a team label but the product area and it is there (below this comment in line 98/109?

gosh89 (approved):

PR #3593, opened 2025-08-04
MTM-64099 Device Isolation by Default

rohitjoshi-c8y:


scmi-c8y (approved):
For some reason I was expecting a much larger change... I feel a bit silly for not reviewing this a week ago now ;-)

BeateRixen (approved):

PR #3614, opened 2025-08-11
Fix broken link and incorrect change types

scmi-c8y:
The link from the second entry back to the first one was incorrect and has been fixed.
I also think it works better to have an "Announcement" followed by an "API change" for a breaking change like this one.
We may also do a second "Announcement" at GA time when the old behaviour is completely removed.
I promise this is the last time I will mess with these announcements!
At least we have a good pattern for the next one now.


content/change-logs/platform-services/mqtt-service-0.9.x-device-isolation-announcement.md
rthrippleton, line 1: As an aside - looking at some other precedents, there's no convention that we actually put the type of the change into the filename, it only serves to make it easier to accidentally break links in situations like this (though this one is fine).

rthrippleton (approved):

kowa-c8y (approved):

schmiel-cumulocity (approved):

content/change-logs/platform-services/mqtt-service-0.9.x-device-isolation-announcement.md
BeateRixen: I agree, there is no need to add the change type to the file name.

BeateRixen (approved):

content/change-logs/platform-services/mqtt-service-0.9.x-device-isolation-announcement.md
scmi-c8y: Agreed - it was only done in this case because there were effectively two change logs for the same change (although I guess the `0.9.x` and `0.9.6` parts would have been sufficient to distinguish them). This has been a learning experience, and I will be more disciplined next time!

PR #3683, opened 2025-09-01
[EDGE] Open Edge tile contents in a new tab

bbyreddy:
Updated the tile layout to include `target="_blank"` attribute for the pages of type "new-tab", so that the tiles open in a new browser tab. This only applies to the "Edge" tile and not the other tiles with external links like "REST API" tile.


content/sector/edge_server/_index.md
rthrippleton, line 1: Is this actually desirable from a user PoV? I come to https://cumulocity.com/docs/ , and
* Click on something like "Platform administration", same window
* Click on "Edge", new window, but it's really just another aspect of the Cumulocity doc

rthrippleton (approved):

BeateRixen:
@carlosceia Can you please check as this is different from what we usually have.

carlosceia:
From a UX perspective, we shouln't do it. Forcing external links to open in a new tab takes control away from users and disrupt the browsing experience. It's always better to let users decide how they want to open links.

BeateRixen (approved):

BeateRixen:
@carlosceia FYI. As discussed, we will go to an external page for Edge only (as this changes the documentation version). On the long run, we should make it clearer that we change the version here and in general in which version we currently are

PR #3684, opened 2025-09-01
Feat/include json-ld

carlosceia:
Include JSON-LD for maximizing documentation’s visibility and usability by search engines and AI bots


scripts/update_dates.py
carlosceia, line 1: @reey and this one, of course

.github/workflows/update-dates.yml
carlosceia, line 1: @reey can you check this? to make sure it doesn't conflict with other files

PR #3695, opened 2025-09-04
update cockpit screenshots

reey:
Only updated those where I noticed a difference compared to the old screenshots


pawel-rynarzewski-c8y (approved):

BeateRixen:
@reey What is the reason for these changes?

reey:
> @reey What is the reason for these changes?

To have the latest versions of our views?

BeateRixen (approved):

PR #3705, opened 2025-09-08
Bugfix/cdh 5384 schema learning with special characters

cheinz-sag:
Hi,

can you pls review?

Thanks


das-c8y (approved):

rthrippleton (changes requested):
Without knowing any of the context, I'm confident in saying that something is very very wrong in this PR, as you didn't mean to change 1428 files :-D

cheinz-sag:
> Without knowing any of the context, I'm confident in saying that something is very very wrong in this PR, as you didn't mean to change 1428 files :-D

Hi Richard, good catch. That would be indeed a bit too much. I took develop by mistake for creating the branch :) Let me fix that

cheinz-sag:
Wrongly generated branch

PR #3717, opened 2025-09-11
Rename Analytics rules plugin to Smart rules 2.0 in shell app docs

nnat-nabu:


content/device-management-application/viewing-device-details.md
```diff
@@ -38,19 +38,20 @@ Details on these additional menu items are provided where required.
 
 The **Alarms** tab provides information on the alarms of a device. See [Working with alarms](/device-management-application/monitoring-and-controlling-devices/#working-with-alarms) for detailed information on alarms.
 
-### Analytics rules {#analytics-rules}
+### Smart rules (NEW) {#smart-rules-new}
```
ben-c8y:
```suggestion
### Smart rules (NEW) {#smart-rules}
```

It's good to use (NEW) in the title, but I don't think we need to use "-new" in the # anchor. Use use `#smart-rules`  so it's easier when we remove the new suffix in future

content/device-management-application/viewing-device-details.md
```diff
@@ -38,19 +38,20 @@ Details on these additional menu items are provided where required.
...
-The **Analytics rules** tab allows to create and manage the Analytics Builder model instances directly from the context of a device or group.
+{{< c8y-admon-preview >}}
+This feature is in Public Preview and may be subject to change in the future.
+{{< /c8y-admon-preview >}}
 
-![Analytics rules tab](/images/streaming-analytics/analytics-rules-plugin/analytics-rules-list.png)
+The **Smart rules (NEW)** tab allows to create and manage the Analytics Builder model instances directly from the context of a device or group.
 
-To create your first rule, refer to [Create your first rule](/streaming-analytics/analytics-rules-plugin/#creating-your-first-analytics-rule).
+![Smart rules new tab](/images/streaming-analytics/analytics-rules-plugin/analytics-rules-list.png)
```
ben-c8y: it shouldn't be called "analytics-rules-plugin". This is another reason why hafing a single PR for the doc will be better. 

There are several references to "analytics-rules-plugin" in both PRs. Once we've decided if we're merging them into one PR, let's merge them and then grep for "analytics-rules" in the entire doc repo to check for remaining errors

ben-c8y: And also check for broken links

content/cockpit/smart-rules-new.md
```diff
@@ -0,0 +1,24 @@
+---
+weight: 80
```
ben-c8y, line 2: We shouldn't have the new smart rules sandwiched in between the 2 topics for the legacy smart rules. I've asked Beate if  we can move the "collection" to a child, but if not we'd need to change the weighting here. I'll let you know the result

PR #3737, opened 2025-09-19
add llms.txt generation step to staging workflow

logu-c8y:


carlosceia:
Maybe we can use the same approach we're using in the website (which is also built with Hugo) and have the llms.txt generated by hugo? here's the output: https://www.cumulocity.com/llms.txt

logu-c8y:
> Maybe we can use the same approach we're using in the website (which is also built with Hugo) and have the llms.txt generated by hugo?


For now, by deleting the local llms.txt file and generating it within the workflow, we can avoid conflicts. If we run into the same problem again, we can look into applying a different approach?

carlosceia:
@logu-c8y created a new PR to generate the llms.txt with hugo, please chek it here: https://github.com/Cumulocity-IoT/c8y-docs/pull/3741

PR #3743, opened 2025-09-22
New glossary page

carlosceia:
Adding the layout and structure of the new glossary page, now all letters are rendered in a single page, but are still handled on separate pages for each letter for easier maintenance.

Added as an example the content for letter A from the glossary document currently under review:

quick link for the preview: https://cumulocity-iot-c8y-docs-build-pr-3743.surge.sh/glossary/


BeateRixen:
I have added the definitions for all terms from the Google Docs, except for two, which are still in discussion (smart rules, subscription). 
As agreed on with @beabehnisch:
- I only added the regular definition for now. The API details, which are not relevant for all terms, must be reviewed by the teams and added in a second step.
- Related terms have been resolved by directly linking the terms within the definitions to other glossary terms. This way, I could not catch all terms in the related terms section since some were not explicitly mentioned in the definition, but a lot and I believe the most important ones. 
- Further links to the respective places in the product documentation have only been added in a few places, we must decide if we want to enhance this in a next step.
@carlosceia Can you please take a final look from your perspective and approve? Do we need to mention that the Search at the top only searches the terms and not the definitions, or is this obvious?

carlosceia:
>   @carlosceia Can you please take a final look from your perspective and approve? Do we need to mention that the Search at the top only searches the terms and not the definitions, or is this obvious?

I made a few improvements to the search, now it's also searching the definitions and the search term is highlighted in the results.

I've also added a link to the glossary on the "additional resources" dropdown in the header

let me know if everything is ok

beabehnisch:
@BeateRixen @carlosceia just realized that the new version leads to many 404s right now - can we redirect all of the single letter URLs like https://cumulocity.com/docs/glossary/a/ to https://cumulocity.com/docs/glossary?

BeateRixen:
> @BeateRixen @carlosceia just realized that the new version leads to many 404s right now - can we redirect all of the single letter URLs like https://cumulocity.com/docs/glossary/a/ to https://cumulocity.com/docs/glossary?

@beabehnisch Where exactly do you see 404s? I do not see any links inside the documentation.

PR #3762, opened 2025-09-30
Update apama-in-c8y-20250829-Cumulocity_Utils_FormatTime.md

skom-c8y:
The change was merged to eu-latest couple of weeks back but automation didn't pick it up


BeateRixen:
@reey Can you please check if there is an issue with the automation?

skom-c8y:
Is the automation based on helm-chart or some other mapping of component vs helm-chart ?

BeateRixen:
> Is the automation based on helm-chart or some other mapping of component vs helm-chart ?

Tristan is on vacation. I will merge this now, but we should check the issue when he is back.

BeateRixen (approved):

PR #3795, opened 2025-10-10
MTM-64874 Fixing old broken links

tsvet:


support-designfitout (commented):
bugfix/MTM-64874/repairing-broken-links

BeateRixen (approved):

PR #3808, opened 2025-10-15
Update Broken link Script

logu-c8y:


carlosceia (approved):

BeateRixen (approved):

broken-links-script/Extractlinks.js
```diff
@@ -63,34 +69,37 @@ const resolveFullUrl = (link, relativePath, fileContent) => {
   }
 
   if (link.startsWith("#")) {
-    const fileDir = path.dirname(relativePath);
+    const fileDir = path.dirname(relativePath).replace(/\\/g, "/");
     const fileName = path.basename(relativePath, ".md");
-    let segments = fileDir.split(path.sep);
+    let segments = fileDir.split("/").filter(Boolean);
```
pawel-rynarzewski-c8y: Could you please explain these changes?

broken-links-script/Extractlinks.js
```diff
@@ -63,34 +69,37 @@ const resolveFullUrl = (link, relativePath, fileContent) => {
...
     }
 
+    // if this file is not rendered, publish from its directory (e.g., /glossary/)
+    const notRendered = hasRenderFalse(fileContent);
+
     let publishedBasePath = "";
-    if (hasBundle) {
+    if (notRendered) {
+      publishedBasePath = segments.join("/");
+    } else if (hasBundle) {
```
pawel-rynarzewski-c8y:
```suggestion
    if (notRendered || hasBundle) {
```

broken-links-script/Extractlinks.js
logu-c8y: I changed this to fix path issues. `path.sep` was causing problems on Windows, and since the workflow runs on another platform, I replaced backslashes with `/` so it works the same everywhere.

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -5,41 +5,59 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
...
+          allFragments = allFragments.concat(collectFragments(frameDoc));
+        }
+      } catch (e) {
+      }
+    }
+
+    allFragments = allFragments.map(f => decodeURIComponent(f).toLowerCase());
+    const normalize = (str) => str.replace(/[^a-z0-9]/gi, '');
+    const exists = allFragments.some(f =>
+      f === decodedFragment || normalize(f) === normalize(decodedFragment)
```
pawel-rynarzewski-c8y: What's the case for normalization, any specific example? Let's say we have an element with id `example-fragment` in the target page and a link with hash `examplefragment`. With the above logic we'll assume the link is correct, while it won't lead user to the right section of the target page. Or am I missing something?

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -5,41 +5,59 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
   const totalTests = urls.length;
 
 
-  const escRegExp = (string) => string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+  const expectFragmentExists = (doc, fragment) => {
+    const decodedFragment = decodeURIComponent(fragment).toLowerCase();
+    const collectFragments = (root) => {
+      const ids = Array.from(root.querySelectorAll('[id]')).map(el => el.id);
+      const names = Array.from(root.querySelectorAll('a[name]')).map(a => a.getAttribute('name'));
+      const hrefs = Array.from(root.querySelectorAll('a[href^="#"]')).map(a => a.getAttribute('href').substring(1));
```
pawel-rynarzewski-c8y: Wouldn't it cause false positive in a case like:

- we have a page with link `<a href="#example-fragment">Link</a>` and a broken target anchor `<a name="example-frament"></a>` on the same page
- the above logic will recognize the source link `<a href="#example-fragment">Link</a>` also as the target, instead of raising issue that there is no anchor with the matching name

Or am I missing something?

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -62,17 +80,44 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
 
       const hasNonHtmlExtension = nonHtmlExtensions.some(ext => url.endsWith(ext));
       const isNonHtmlResource = hasNonHtmlExtension || url.includes('/files/') || url.includes('/downloads/');
+      const isPrivateGithubrepository = url.includes('github.com/Cumulocity-IoT/');
```
pawel-rynarzewski-c8y: Do we link only to private repos in this org? How many such links? Maybe we should avoid them in the public docs, as general public won't be able to access them anyway?

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: > Wouldn't it cause false positive in a case like:
> 
> * we have a page with link `<a href="#example-fragment">Link</a>` and a broken target anchor `<a name="example-frament"></a>` on the same page
> * the above logic will recognize the source link `<a href="#example-fragment">Link</a>` also as the target, instead of raising issue that there is no anchor with the matching name
> 
> Or am I missing something?

yes, you are correct. Including a[href="#"] in the collected fragments can cause false positives, since those represent source links, not actual fragment targets.

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: I was trying to fix this link: https://javadoc.io/doc/io.rest-assured/rest-assured/3.0.0/io/restassured/specification/RequestSpecification.html#content-byte:A- in this link if we use normalize then only it is working. You're right, the normalization could cause false positives in that case. I’ve removed it so the check now matches fragments exactly as they appear.

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: There are around 46 links in total. I have checked each of them, and all are public repositories. I added this isPrivateGithubrepository check because we are using this in c8y-docs-ops, which is a private repository. I will remove this condition.

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -5,41 +5,54 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
   const totalTests = urls.length;
 
 
-  const escRegExp = (string) => string.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+  const expectFragmentExists = (doc, fragment) => {
+    const decodedFragment = decodeURIComponent(fragment).toLowerCase();
```
pawel-rynarzewski-c8y: Why `.toLowerCase()`? You need to make exact case-sensitive comparisons, because if you have a link https://cumulocity.com/docs/get-familiar-with-the-ui/gui-features/#main-Screen-elements and target element is `<h3 id="main-screen-elements">...</h3>`, user won't be correctly redirected to that element in browser, but your script will recognize it as matching and won't raise an issue.

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -5,41 +5,54 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
...
+    const iframes = doc.querySelectorAll('iframe, frame');
+    for (const frame of iframes) {
+      try {
+        const frameDoc = frame.contentDocument || frame.contentWindow?.document;
+        if (frameDoc) {
+          allFragments = allFragments.concat(collectFragments(frameDoc));
+        }
+      } catch (e) {
+      }
+    }
```
pawel-rynarzewski-c8y, line 26: Can you give me example of the case you're addressing here? I've made some tests and if you have a document with iframe, and iframe contains an anchor, you cannot use url like `/main-document.html#a-name-from-iframe` to link to the anchor inside the iframe, i.e. user won't be scrolled to the right position within the iframe.

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -5,41 +5,54 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
...
-        .map(a => a.getAttribute('href'))
-        .filter(href => href && href.startsWith('#'));
-      cy.log(`Available fragments on page:\n${allFragments.join('\n')}`);
+      const anchors = Array.from(doc.querySelectorAll('a[href^="#"]')).map(a => a.getAttribute('href'));
+      const ids = Array.from(doc.querySelectorAll('[id]')).map(el => el.id);
+      cy.log(`Available GitHub anchors:\n${anchors.join('\n')}`);
+      cy.log(`Available GitHub IDs:\n${ids.join('\n')}`);
+      const exists = anchors.some(href => href === `#${fragment}`) || ids.some(id => id === fragment);
+      expect(exists, `Fragment "#${fragment}" should exist in GitHub page`).to.be.true;
     });
```
pawel-rynarzewski-c8y: Actually, `<a href="#fragment">` is a regular link. What's not standard on GitHub pages is that they prefix target `id`s with `user-content-`, yet you can still use `#fragment` without prefix to link to such an element. So, if you'd like to gather all `id`s in their linkable form, then it'd be like:
```
const ids = Array.from(doc.querySelectorAll('[id]')).map(el => el.id.replace(/^user-content-/, ''));
```
and there should be no need to query for `a[href]`.

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -62,17 +75,37 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
...
+          cy.request({
+            url: `https://registry.npmjs.org/${pkg}`,
+            failOnStatusCode: false,
+            headers: { Accept: 'application/vnd.npm.install-v1+json' }
+          }).then((res) => {
+            expect(res.status, `npm registry status for ${pkg}`).to.eq(200);
+          });
+        }
+        cy.visit(encodedUrl, { timeout: 50000, failOnStatusCode: false });
+        cy.url().should('include', '/package/%40');
```
pawel-rynarzewski-c8y: Can you explain this logic? I see you match only packages with scoped names starting with `@` and currently we only have such urls in docs, but I wouldn't make it that specific. Could we assume that anything after `https://www.npmjs.com/package/` is a full package name?
I also read about constraints on accessing `https://www.npmjs.com` by bots and that we need to use registry instead. But what's the purpose of then encoding url, visiting it and checking for `%40`?

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -117,12 +150,23 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
...
-        cy.visit(url, { timeout: 20000 });
-        cy.document().its('body').should('not.be.empty');
+        cy.request({
+          url: url,
+          failOnStatusCode: false
+        }).then((response) => {
+          const contentType = response.headers['content-type'] || '';
+          if (!contentType.includes('text/html')) {
+            cy.log(`Non-HTML content detected for ${url}, skipping cy.visit()`);
+            expect(response.status).to.be.oneOf([200, 201, 202, 203, 204, 301, 302, 304]);
```
pawel-rynarzewski-c8y, line 160: Should we check that `response.body` is not empty? If we direct user there, there should be some content. Do you have any example of what non-html resource might that be?

broken-links-script/Extractlinks.js
```diff
@@ -50,6 +49,13 @@ const shortcodeMapping = {
   "email-c8y-info": "info@cumulocity.com",
 };
 
+const hasRenderFalse = (fileContent) => {
+  const fm = fileContent.match(/(^|\n)---\s*[\s\S]*?\n---/);
+  if (!fm) return false;
+  const block = fm[0];
+  return /_build:\s*[\r\n]+[\s\S]*?render:\s*false\b/i.test(block);
```
pawel-rynarzewski-c8y: Here you check the contents of the front matter, right?
```suggestion
  const frontMatter = fm[0];
  return /_build:\s*[\r\n]+[\s\S]*?render:\s*false\b/i.test(frontMatter);
```

broken-links-script/Extractlinks.js
```diff
@@ -63,34 +69,35 @@ const resolveFullUrl = (link, relativePath, fileContent) => {
...
+    // if this file is not rendered, publish from its directory (e.g., /glossary/)
+    const notRendered = hasRenderFalse(fileContent);
+
     let publishedBasePath = "";
-    if (hasBundle) {
+    if (notRendered || hasBundle) {
       publishedBasePath = segments.join("/");
     } else {
-      publishedBasePath = fileName === "index" ? fileDir : path.join(fileDir, fileName);
+      publishedBasePath = fileName === "index" ? fileDir : path.join(fileDir, fileName).replace(/\\/g, "/");
```
pawel-rynarzewski-c8y: I'd say don't use `path.join` here, it's for joining path parts with platform specific separator which is not what you need here, just do: ``` `${fileDir}/${fileName}` ``` or `[fileDir, fileName].join('/')`.

broken-links-script/Extractlinks.js
```diff
@@ -63,34 +69,35 @@ const resolveFullUrl = (link, relativePath, fileContent) => {
   }
 
   if (link.startsWith("#")) {
-    const fileDir = path.dirname(relativePath);
+    const fileDir = path.dirname(relativePath).replace(/\\/g, "/");
```
pawel-rynarzewski-c8y: If you need to convert disk path (platfom specific) into url path, you could do:
```suggestion
    const fileDir = path.dirname(relativePath).replaceAll(path.sep, '/');
```

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: removed .toLowerCase() in the script for exact case-sensitive.

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: The intention here isn’t to support direct navigation to iframe anchors via #fragment, but to verify that referenced anchors (even inside embedded documents) actually exist. We have some pages that load documentation or content in iframes, so this logic helps our validation detect missing anchors there too.

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: I have updated the checkGithubFragment function to strip the prefix and rely only on IDs.

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: Currently, I am matching only scoped packages (@scope/pkg) since those are the only ones we have in our docs. The registry call is used to confirm that the package actually exists (to avoid npmjs.com’s bot-blocks).
After that, the encoded URL visit (%40) ensures the page resolves correctly for scoped packages, since npmjs.com automatically redirects them to the encoded form

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: For example, Link: https://download.cumulocity.com/Apama/Debian/. These are file repositories or download links, not HTML pages, so I only check that they return a valid status and skip checking the body.

broken-links-script/Extractlinks.js
logu-c8y: I replaced `path.join` with a simple string join like `${fileDir}/${fileName}` to avoid platform-specific separators

broken-links-script/cypress/e2e/link-checker.cy.js
pawel-rynarzewski-c8y: Would be easier if you provided an example link and use case. Sure, with the above mechanism we can verify that the anchor exists inside the iframe on the target page (if they are from the same domain). My concern is also user experience in case we have a link in our docs which refers to a specific anchor inside an iframe on another page (internal or external), then user won't be navigated to the referred section directly, but just to the top of the page.

broken-links-script/cypress/e2e/link-checker.cy.js
pawel-rynarzewski-c8y: > Currently, I am matching only scoped packages (@scope/pkg) since those are the only ones we have in our docs. The registry call is used to confirm that the package actually exists (to avoid npmjs.com’s bot-blocks).

ok, that's clear

> After that, the encoded URL visit (%40) ensures the page resolves correctly for scoped packages, since npmjs.com automatically redirects them to the encoded form

Sorry, didn't get that. If you navigate to the encoded URL, then it's expected that you detect `%40` coming from the encoded URL. It does not prove that npmjs.com resolved anything or redirected you.
This check will also pass for a made-up package name:
```ts
cy.visit('https://www.npmjs.com/package/%40c8y/made-up-package', { timeout: 50000, failOnStatusCode: false });
cy.url().should('include', `/package/%40`);
```
Isn't the registry check enough?

broken-links-script/cypress/e2e/link-checker.cy.js
pawel-rynarzewski-c8y: This link seems to be special as it actually returns 2 `content-type` headers, but `response.headers` only shows you one of them.



Anyway, I'd consider checking `response.body` to not be empty, just as you check detected `text/html` pages for non-empty HTML body.

```suggestion
            expect(response.status).to.be.oneOf([200, 201, 202, 203, 204, 301, 302, 304]);
            expect(response.body).not.to.be.empty;
```

broken-links-script/Extractlinks.js
```diff
@@ -63,34 +69,35 @@ const resolveFullUrl = (link, relativePath, fileContent) => {
...
+      publishedBasePath = fileName === "index" ? fileDir : `${fileDir}/${fileName}`;
     }
-    publishedBasePath = publishedBasePath
-      .replace(/\\/g, "/")
-      .replace(/^\/+/, "")
-      .replace(/\/+$/, "");
-    return `${BASE_URL.replace(/\/$/, "")}/${publishedBasePath}/#${link.substring(1)}`;
+    publishedBasePath = publishedBasePath.replace(/^\/+|\/+$/g, "");
+    const base = `${BASE_URL.replace(/\/$/, "")}/${publishedBasePath ? publishedBasePath + "/" : ""}`;
+    return `${base}#${link.substring(1)}`;
```
pawel-rynarzewski-c8y: The replacements are to prevent excessive slashes, right? What if you just concatenated the parts first, and then dropped any unneeded slashes except for the protocol part, e.g.:
```ts
function test(BASE_URL, publishedBasePath, link) {
  let url = `${BASE_URL}/${publishedBasePath}#${link.substring(1)}`;
  // remove duplicated slashes except for protocol part
  url = url.replace(/([^:]\/)\/+/g, '$1');
  // remove trailing slash before hash
  url = url.replace(/\/#/g, '#');
  return url;
}

console.log(test('https://example.com/', '/docs/ui/', '#my-section'));
console.log(test('http://example.com/', 'docs/ui', '#my-section'));
```

broken-links-script/Extractlinks.js
```diff
@@ -50,6 +49,13 @@ const shortcodeMapping = {
   "email-c8y-info": "info@cumulocity.com",
 };
 
+const hasRenderFalse = (fileContent) => {
+  const fm = fileContent.match(/(^|\n)---\s*[\s\S]*?\n---/);
+  if (!fm) return false;
+  const frontMatter = fm[0];
+  return /_build:\s*[\r\n]+[\s\S]*?render:\s*false\b/i.test(frontMatter);
+};
```
pawel-rynarzewski-c8y, line 60: We already use `gray-matter` in other parts of `c8y-docs` to read/write frontmatters, might be easier and more reliable than regex.

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -5,41 +5,52 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
...
+      }
+    }
+
+    const exists = allFragments.some(f => f === decodedFragment);
+
+    if (!exists) {
+      cy.log(`Available fragments (including frames):\n${allFragments.join('\n')}`);
+    }
+
+    expect(exists, `An element with id or name = "#${fragment}" should exist in HTML or frames`).to.be.true;
```
pawel-rynarzewski-c8y:
```suggestion
    expect(exists, `An element with id or name = "${fragment}" should exist in HTML or frames`).to.be.true;
```

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -5,41 +5,52 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
...
-      
-      expect(anchorExists, `Fragment "#${normalizedFragment}" should exist in href attribute of an <a> tag`).to.be.true;
-      const allFragments = Array.from(doc.querySelectorAll('a'))
-        .map(a => a.getAttribute('href'))
-        .filter(href => href && href.startsWith('#'));
-      cy.log(`Available fragments on page:\n${allFragments.join('\n')}`);
+      const ids = Array.from(doc.querySelectorAll('[id]')).map(el => el.id.replace(/^user-content-/, ''));
+      cy.log(`Available GitHub IDs:\n${ids.join('\n')}`);
+      const exists = ids.some(id => id === fragment);
+      expect(exists, `Fragment "#${fragment}" should exist in GitHub page`).to.be.true;
```
pawel-rynarzewski-c8y:
```suggestion
      expect(exists, `Element with id "user-content-${fragment}" should exist in GitHub page`).to.be.true;
```

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: We have some documentation pages that embed other HTML files inside iframes example: https://javadoc.io/doc/io.rest-assured/rest-assured/3.0.0/io/restassured/specification/RequestSpecification.html#content-byte:A- . These embedded pages contain anchors (like method or section references). Our validation needs to confirm that these anchors actually exist inside the iframe content, even though users can’t directly navigate to them using a URL fragment. Here with this example link: when we open this link first time, we are redirecting to the fragment. But If we moves to somewhere in the page and try to open the URL, it won't redirect to the specific fragment.

broken-links-script/cypress/e2e/link-checker.cy.js
logu-c8y: Yeah, the registry check alone is enough to confirm the package exists. I’ll remove the extra encoded URL visit step and keep only the registry validation for scoped packages.

broken-links-script/Extractlinks.js
logu-c8y: I updated the code to concatenate the parts first and then clean up duplicate slashes as you suggested.

broken-links-script/Extractlinks.js
logu-c8y: I update the code to use gray-matter instead of regex

broken-links-script/cypress/e2e/link-checker.cy.js
pawel-rynarzewski-c8y: I see, in this case the target external page takes care of passing the hash forward to the iframe:
```js
var hash = window.location.hash;
if (hash) {
    var iframe = $('.fill iframe');
    iframe.attr("src", iframe.attr('src') + hash);
}
```
so the user is redirected at least on first load.
Got your point, so we need to gather these anchors as well. :+1:

broken-links-script/Extractlinks.js
```diff
@@ -63,34 +72,36 @@ const resolveFullUrl = (link, relativePath, fileContent) => {
...
+      publishedBasePath = fileName === "index" ? fileDir : `${fileDir}/${fileName}`;
     }
-    publishedBasePath = publishedBasePath
-      .replace(/\\/g, "/")
-      .replace(/^\/+/, "")
-      .replace(/\/+$/, "");
-    return `${BASE_URL.replace(/\/$/, "")}/${publishedBasePath}/#${link.substring(1)}`;
+    let url = `${BASE_URL}/${publishedBasePath ? publishedBasePath + "/" : ""}#${link.substring(1)}`;
+    url = url.replace(/([^:]\/)\/+/g, '$1');
+    url = url.replace(/\/#/g, '#');
```
pawel-rynarzewski-c8y: You don't need to check if `publishedBasePath` is not an empty string, because concatenating empty string does not change the string. If you use regex, let's keep a short comment what it does.
```suggestion
    let url = `${BASE_URL}/${publishedBasePath}#${link.substring(1)}`;
    url = url.replace(/([^:]\/)\/+/g, '$1'); // removes duplicate slashes
    url = url.replace(/\/#/g, '#'); // removes slash before hash
```

And if you want to keep slash before hash:
```suggestion
    let url = `${BASE_URL}/${publishedBasePath}/#${link.substring(1)}`;
    url = url.replace(/([^:]\/)\/+/g, '$1'); // removes duplicate slashes
```
Note that if `publishedBasePath` is empty, then `//` will be removed by the regex replacement.

```
function test1(BASE_URL, publishedBasePath, link) {
  let url = `${BASE_URL}/${publishedBasePath}#${link.substring(1)}`;
  url = url.replace(/([^:]\/)\/+/g, '$1'); // removes duplicate slashes
  url = url.replace(/\/#/g, '#'); // removes slash before hash
  return url;
}

function test2(BASE_URL, publishedBasePath, link) {
  let url = `${BASE_URL}/${publishedBasePath}/#${link.substring(1)}`;
  url = url.replace(/([^:]\/)\/+/g, '$1'); // removes duplicate slashes
  return url;
}

console.log(test1('https://example.com/', '/docs/2025/test/', '#section-name'));
> https://example.com/docs/2025/test#section-name
console.log(test2('https://example.com/', '/docs/2025/test/', '#section-name'));
> https://example.com/docs/2025/test/#section-name
console.log(test1('https://example.com/', '', '#section-name'));
> https://example.com#section-name
console.log(test2('https://example.com/', '', '#section-name'));
> https://example.com/#section-name
```

PR #3884, opened 2025-11-05
Example

neuerburg:
Just a test


BeateRixen:
> Just a test

Should we set the PR to draft?

jellecumulocity:
> > Just a test
> 
> Should we set the PR to draft?

I don't mind what status this takes. But just consider that it's not ready to be published.

PR #3940, opened 2025-11-21
[MTM-65011] Add new audit log for CA feature, extend good practice wh…

mbak-c8y:
…en OTP password has been compromised

### DIFF
--- a/content/device-certificate-authentication/device-certificates.md
+++ b/content/device-certificate-authentication/device-certificates.md
@@ -435,22 +435,41 @@ If you suspect or confirm that a device certificate has been compromised, follow
 
 Enrollment processes may be compromised when insecure practices are used, for example:
 - Using a device ID-derived one-time password (OTP)
-- Using the same OTP across all enrollment requests
+- Using the same OTP across multiple enrollment requests
 
-If such a situation occurs, proceed with the following steps:
+If such a situation occurs, follow these steps to mitigate the risk:
 
 1. **Disable auto-registration on the tenant’s CA certificate**
-    - This stops any new automatic registrations immediately.
-    - Notice that existing registered devices will remain connected and operational.
+    - This immediately stops any new automatic registrations.
+    - Existing registered devices will remain connected and operational.
+
 2. **Communicate and remediate unsafe procedures**
     - Inform all stakeholders about the issue.
     - If necessary, disable users who were following unsafe enrollment practices.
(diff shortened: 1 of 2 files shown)


kowa-c8y (approved):

content/device-certificate-authentication/device-certificates.md
```diff
@@ -435,22 +435,41 @@ If you suspect or confirm that a device certificate has been compromised, follow
...
+            - **Activity:** `Tenant certificate authority (CA) signed certificate for device`
+            - **Type:** `TenantCertificateAuthority`
+            - **Text:** containing *`Certificate serial number hex: '%s'`*
+        2. **Verify the signer**  
+           Confirm whether the certificate was signed by an authorized device.
+            - If it **was not** signed by an authorized user, it may indicate fraudulent or unintended device registration.
+        3. **Revoke unauthorized certificates**  
+           Use the [certificate revocation list](/device-certificate-authentication/managing-trusted-certificate-settings/#crl-settings) to revoke any certificates that were not properly authorized. This prevents unauthorized devices from being accepted as trusted.
+
+4. **Re-enable auto-registration**
```
BeateRixen, line 468: I assume this has to be indented as well?

jkot-c8y: not sure. Because there is also 1. 2. 3. in the same indentation level (see line 451). 
@mbak-c8y maybe some other markers for this embedded enumeration, eg. (1), (2), (3), or a), b), c) ?

BeateRixen: I just checked in the preview. You are right. Actually, I would remove the numbering completely. You could use bullets instead but since the action points are already formatted in bold this is not mandatory. If you use bullets, you should remove the bullet from line 464 in the second step. It anyway does not make much sense as long as no other entries follow.

mbak-c8y: Screenshot from 2025-11-25 12-42-52
Does it look better?

BeateRixen: In general, the layout/structure could be clearer, but let´s leave it for now, and after this PR is merged, I can take a look.

BeateRixen (approved):

content/device-certificate-authentication/device-certificates.md
mbak-c8y: Thanks

PR #3944, opened 2025-11-24
CIE-5505 - Having DataHub Edge doc in a single place

rthrippleton:
Somebody complained (reasonably) that the documentation for DataHub on Edge was slightly fragmented, and would sometimes send you on a link wild goose chase between the DataHub and Edge sections of the doc.

I've transplanted everything into the Edge section of the doc. I think this makes more sense now that Edge is just a Kubernetes-based mini-cloud, and DataHub is "just" another component (with some caveats).

For the same reasons, I've removed a bunch of the kubernetes diagnostic content, because Edge health checks all of its components, and aggregates all of its logs.

PR #3992, opened 2025-12-08
Fix broken links between CDH and Edge doc

rthrippleton:
#3944 changed/removed a whole bunch of CDH + Edge doc, mostly removing CDH + Edge topics from the CDH doc. This is sort-of a forward port of that change.


BeateRixen:
Not sure if this is required but I assume we should also remove it from here: https://github.com/Cumulocity-IoT/c8y-docs/blob/develop/static/llms.txt

rthrippleton:
> Not sure if this is required but I assume we should also remove it from here: https://github.com/Cumulocity-IoT/c8y-docs/blob/develop/static/llms.txt

Looking at that I think it's auto-generated, surely?

BeateRixen:
> > Not sure if this is required but I assume we should also remove it from here: https://github.com/Cumulocity-IoT/c8y-docs/blob/develop/static/llms.txt
> 
> Looking at that I think it's auto-generated, surely?

Is this the case @carlosceia?

BeateRixen (approved):

cheinz-sag (approved):

PR #4045, opened 2026-01-07
Update mqtt-examples.md

BeateRixen:
Adapted the text as one link does no longer exist.

### DIFF
--- a/content/device-integration/mqtt-bundle/mqtt-examples.md
+++ b/content/device-integration/mqtt-bundle/mqtt-examples.md
@@ -1109,7 +1109,7 @@ What does the script do?
 -   Update the device hardware information by putting a `"S123456789"` serial, a `"MQTT test model"` model and a `"Rev0.1"` revision.
 -   Subscribe to the static operation templates for the device – this will result in an ``on_message`` method call every time a new operation is created.
 -   Start the ``device_loop_thread`` which sends a temperature measurement every 7 seconds.
--   Prepare a ``task_queue``, used mainly to overcome deadlock [issue](https://github.com/eclipse/paho.mqtt.python/issues/354) in Python Paho library. Task queue will run all tasks one by one.
+-   Prepare a ``task_queue``, which runs all tasks one by one.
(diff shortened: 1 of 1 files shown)


content/device-integration/mqtt-bundle/mqtt-examples.md
```diff
@@ -1109,7 +1109,7 @@ What does the script do?
 -   Update the device hardware information by putting a `"S123456789"` serial, a `"MQTT test model"` model and a `"Rev0.1"` revision.
 -   Subscribe to the static operation templates for the device – this will result in an ``on_message`` method call every time a new operation is created.
 -   Start the ``device_loop_thread`` which sends a temperature measurement every 7 seconds.
--   Prepare a ``task_queue``, used mainly to overcome deadlock [issue](https://github.com/eclipse/paho.mqtt.python/issues/354) in Python Paho library. Task queue will run all tasks one by one.
+-   Prepare a ``task_queue``, which runs all tasks one by one.
```
BeateRixen, line 1112: It seems that the issue mentioned in the MQTT example does no longer exist. At least the link is broken and I can´t find it somewhere else. Should we still leave this step (prepare a task_queue) or remove the entire last step? Not sure what "used mainly" means here.

BeateRixen:
@scmi-c8y @JanePorter Could you please check/approve this change? Or could you add others as reviewers if needed?

content/device-integration/mqtt-bundle/mqtt-examples.md
scmi-c8y: The link (https://github.com/eclipse-paho/paho.mqtt.python/issues/354) works for me. However, the issue is closed and seems to have been fixed ~5 years ago. I suspect the sample could be re-written without using a task queue - I've used Paho Python a lot and never needed to do this! That would need to be raised as a separate issue. For now, the code should still work as written so I think this change is fine.

scmi-c8y (approved):

PR #4048, opened 2026-01-07
MTM-64963 MQTT keywords in lowercase, C8Y-style.

scmi-c8y:


scmi-c8y:
Decided to go with the approach in #4047 instead.

PR #4106, opened 2026-01-20
Changelog entry for apama mqtt service EPL API removal [PAM-35340]

skom-c8y:


matj-sag (approved):

content/change-logs/analytics/apama-in-c8y-20260120-removed-mqtt-service.md
```diff
@@ -0,0 +1,18 @@
+---
+date: 
+title: Removal of MQTT Service support in EPL Apps
+change_type:
+  - value: change-inv-3bw8e
+    label: Announcement
```
BeateRixen, line 6: This is actually not an announcement as it describes something that has already been implemented. So I would rather use Improvement here.

matj-sag: Is removing something really an Improvement? I definitely consider 'we removed this' as much of an annoucement as 'we added this'

BeateRixen: > Is removing something really an Improvement? I definitely consider 'we removed this' as much of an annoucement as 'we added this'

This way, each change log could actually be seen as an announcement for the change, be it a fix, a feature, or anything else.  We use the term announcement a bit more strictly, announcing important upcoming changes upfront that might require user action.

matj-sag: Oh, I see, so we announce that we *will* remove it, in the deprecation notice, but use improvement for the thing we *did* in removing it

skom-c8y: I actually went with 'Announcement' because that's how we say removals are documented

image

https://github.com/Cumulocity-IoT/c8y-docs/blob/95c350afe16a5746aba24de8c5eb2103a33539a6/content/change-logs/_index.md?plain=1#L15

BeateRixen: Actually, it is meant to announce them upfront. @vaillant @neuerburg Since you were involved in the original categorization, what is your take here? Is the removal of a feature that previously has been announced as deprecated also an "announcement"? I used to handle it as an "Improvement". Since this seems not to be clear, I would like to specify it in the release notes creation document.

neuerburg: In general, I do agree that removals should be handled as announcements (especially if they are relevant changes that need to be announced ahead of time according to our compatibility policy).

In this case, I believe the announcement is not clear to many readers. I understood from @niva-c8y that here "mqttservice" is not referring to the MQTT Service in development but to our old MQTT endpoint. This should be explained.

Also, it is not clear which action is required by the customer. If that was explained in the original deprecation note, a good practice would be to link it.

In general, it is also important to explain that the feature has been deprecated previously.

skom-c8y: Ok, updated the change log based on review comments. specifically called out the distinction between `tenant isolation` (removed capability) vs `device isolation` (new capability)

Left the label as `Announcement` as suggested

matj-sag (approved):

BeateRixen (approved):

PR #4138, opened 2026-01-27
Update extract links file to get the links with localhost and update …

logu-c8y:
…hugo setup in github workflow

### DIFF
--- a/broken-links-script/Extractlinks.js
+++ b/broken-links-script/Extractlinks.js
@@ -16,7 +16,7 @@ const getMarkdownFiles = (dir) => {
   return markdownFiles;
 };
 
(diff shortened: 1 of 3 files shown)


broken-links-script/Extractlinks.js
```diff
@@ -16,7 +16,7 @@ const getMarkdownFiles = (dir) => {
   return markdownFiles;
 };
 
-const BASE_URL = "https://cumulocity.com/docs";
+const BASE_URL = "http://localhost:1313/docs/";
```
igre-c8y, line 19: should it be aligned with that we had previously, or ?

```suggestion
const BASE_URL = "http://localhost:1313/docs";
```

PR #4186, opened 2026-02-03
Testing GitHub AI suggestions

logu-c8y:


BeateRixen:
@logu-c8y What are the changes in the json files about?

logu-c8y:
> @logu-c8y What are the changes in the json files about?

I deleted those files and moved the dependency installation into the workflow instead. The script uses JSON5 rather than the built-in JSON parser because the AI output isn't always perfectly valid JSON, and JSON5 is a bit more forgiving with small formatting issues.

BeateRixen (approved):

PR #4214, opened 2026-02-05
[MTM-65134] added change log for stuff moved to GA, added change-log …

janhommes:
…agent

you can  now use an change-log agent to create the release note:
image

Then you only need to provide the data and a file is created for you:

image


content/change-logs/application-enablement-solutions/cumulocity-2025.373.0-features-moved-to-ga.md
```diff
@@ -0,0 +1,17 @@
...
+  - value: change-inv-3bw8e
+    label: Announcement
+component:
+  - value: component-YdSEScrEC
+    label: Cockpit
+build_artifact:
+  - value: tc-pjJiURv9Y
+    label: ui-c8y
+ticket: MTM-65134
+version: 2025.373.0
```
janhommes: This was change in this helm version, but actually is visible in cockpit. What should we put here or in the build-arifact?

PR #4439, opened 2026-03-17
create Edge card and update index.html file to display Edge PDF button

logu-c8y:


bbyreddy:
@logu-c8y , this change seems to have distorted the tiles. The new tiles look smaller and different (without margins). Is that intentional?

content/edge-kubernetes-card.md
```diff
@@ -0,0 +1,10 @@
+---
+title: Edge
+bundlefolder: edge-kubernetes
+icon: "c8y-icon c8y-icon-notification"
+type: root
+layout: redirect
+sector:
+  - edge_server
+weight: 50
+---
```
bbyreddy, line 10: Not sure why this card is required as the Edge tiles were showing normally without this as well. Because of this, now we see multiple "Edge" dropdowns in the Edge navigation pane.

logu-c8y:
> @logu-c8y , this change seems to have distorted the tiles. The new tiles look smaller and different (without margins). Is that intentional?

I am adding a feature to generate and download the PDFs as a ZIP file from the main tiles. While integrating the download buttons, the tile layout got slightly distorted. I am working on fixing the layout so it matches the original design.

content/edge-kubernetes-card.md
logu-c8y: In `Edge`, when we click on a tile, the content opens directly, unlike other tiles which have `sub-tiles`. I am working on adding a download feature for the main tiles. For `Edge`, this will be a `PDF download`, while for the other tiles it will be a `ZIP download.` The sub-tiles already have a `PDF download` option.

To generate the `PDF` for `Edge`, we need an `edge-kubernetes-card`. The PDF generator relies on this card to identify the content and create the PDF, that is the reason I created `edge-kubernetes-card`.

BeateRixen (approved):

carlosceia (approved):

PR #4440, opened 2026-03-17
Added a preview box

BeateRixen:
The preview information is currently a bit hidden. I added the standard preview box.


skom-c8y (approved):

skom-c8y (changes requested):
Actually, this part of the doc is auto-generated as part of the api doc build, and the same sources are also used to render the doc in the web application.

It will not be possible to include the admon directly in the sources. I propose we close this

BeateRixen:
> Actually, this part of the doc is auto-generated as part of the api doc build, and the same sources are also used to render the doc in the web application.
> 
> It will not be possible to include the admon directly in the sources. I propose we close this



> Actually, this part of the doc is auto-generated as part of the api doc build, and the same sources are also used to render the doc in the web application.
> 
> It will not be possible to include the admon directly in the sources. I propose we close this

Ok, got it, will close the PR

PR #4488, opened 2026-03-31
Extract correct glossary URLs and exclude screenshots and added README.md file

logu-c8y:


BeateRixen (approved):

broken-links-script/Extractlinks.js
```diff
@@ -85,18 +86,21 @@ const resolveFullUrl = (link, relativePath, fileContent) => {
       }
     }
 
-    // if this file is not rendered, publish from its directory (e.g., /glossary/)
```
pawel-rynarzewski-c8y, line 88: This comment suggests that glossary was already considered as non-renderable. Does it just not have a `render` flag at all? Maybe `return data?._build?.render === false;` condition is to strict and we should check for `false` or `undefined`?

broken-links-script/Extractlinks.js
```diff
@@ -85,18 +86,21 @@ const resolveFullUrl = (link, relativePath, fileContent) => {
...
+    const notRendered =
+      hasRenderFalse(fileContent) || fileDir === "glossary";
 
     let publishedBasePath = "";
-    if (notRendered || hasBundle) {
+
+    if (hasBundle) {
       publishedBasePath = segments.join("/");
+    } else if (notRendered) {
+      publishedBasePath = fileDir;
```
pawel-rynarzewski-c8y: This may potentially affect links other than glossary ones, are you sure it's not breaking them?

broken-links-script/Extractlinks.js
pawel-rynarzewski-c8y: `hasRenderFalse` checks only `data?._build?.render`, but glossary entries are marked with `data?.build?.render`, if we extend the condition to take this into account, then all glossary links (80) are passing.

pawel-rynarzewski-c8y (approved):

PR #4545, opened 2026-04-10
Add prerequisites warning for Edge upgrade

raghuvamsi-d:
Added caution note regarding prerequisites for upgrading Edge.

### DIFF
--- a/content/edge/manage-edge-bundle/version-upgrade.md
+++ b/content/edge/manage-edge-bundle/version-upgrade.md
@@ -3,6 +3,11 @@ weight: 15
 title: Upgrading Edge
 layout: redirect
 ---
+{{< c8y-admon-caution >}}
+Before upgrading your Edge environment, ensure that all prerequisites are met.
+<br>See [Prerequisites](/edge/installing-edge/#prerequisites) for the supported Kubernetes versions and requirements.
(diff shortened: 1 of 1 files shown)


content/edge/manage-edge-bundle/version-upgrade.md
```diff
@@ -3,6 +3,11 @@ weight: 15
 title: Upgrading Edge
 layout: redirect
 ---
+{{< c8y-admon-caution >}}
```
rthrippleton, line 6: As we only change k8s versions and do significant resource changes in major versions, can we suggest that this is only relevant for such upgrades? Don't want to overly worry somebody doing a routine fix update.

raghuvamsi-d: I thought of keeping this contextual as you suggested here @rthrippleton, but I felt better to be cautious during any upgrade and kept it very generic.

rthrippleton: I guess instead of saying "don't worry", we could add context - your generic wording plus "most likely to change significantly during a major release".

rthrippleton (approved):

bbyreddy (approved):

PR #4556, opened 2026-04-14
docs: [CDH-6400] Update docs for writing inventory entries to CDC tables

flsc-c8y:


content/datahub/streaming-lake-ingestion-bundle/using.md
```diff
@@ -186,7 +186,20 @@ The "latest data" versions of the tables reflect this update:
 
 ##### Deleting inventory data
 
-Currently, delete operations are not processed.
+Currently, delete operations of inventory entries are only written for CDC tables. The written records only contain the 
+fields `id`, `eventType` and `lastUpdated`, where the latter one is set to the publishing time of the event. 
+All other values will be `null`. Fragment tables are not being updated for delete operations.
```
eickler: I didn't uderstand that part: First "delete operations of inventory entries are only written for CDC tables", then "Fragment tables are not being updated for delete operations". What is the meaning of this? Does it mean that you see the MANAGED_OBJECT_DELETE only in the table "cdc_inventory.inventory"? If that is the case, please just write "When a managed object is deleted, an update is written to cdc_inventory.inventory like this: … <example> … Only the key columns are updated."

Then please update the limitations in the "Current preview limitations" section below. Remove "Delete events are not processed." and replace with "Delete events are only recorded for the inventory change data capture, not for the latest inventory." (or similar)

eickler (changes requested):

jaro-herod (approved):

eickler (approved):

PR #4592, opened 2026-04-20
docs: [CDH-6413] Add section for integration with query engines

flsc-c8y:


content/datahub/streaming-lake-ingestion-bundle/using.md
```diff
@@ -793,6 +793,11 @@ The following data is moved to the `trash` table.
 }
 ```
 
+## Integration with query engines {#integration-with-query-engines}
+
+As Streaming Lake Ingestion stores data in the Apache Iceberg format, you can integrate it with a variety of query engine that support Apache Iceberg, such as Dremio, Trino, Spark, Databricks, etc.
```
flsc-c8y: I don't know if Spark is a real "query engine", but I for now left it in the list.

eickler (changes requested):
Please check https://cumulocity-iot-c8y-docs-build-pr-4592.surge.sh/datahub/streaming-lake-ingestion/#using-datahub-query-to-analyze-lake-data I think the only new information that is conveyed by the PR is that for Dremio, users automatically get query permissions assigned. Is that right? 

In that case, we could just change the first sentence of the "Using DataHub Query" section linked above to

> If your tenant includes a subscription to DataHub Query (Dremio), the system automatically sets up a data source with your tenant ID for querying the lake and grants all your users read permissions to the lake.

Missing is the persistent assignment. We could add to the limitations section that 

> Users automatically get read permissions to the Dremio data source assigned. These read permissions cannot be manually revoked.

eickler (approved):

PR #4767, opened 2026-05-27
CIE-5604 - configuring Edge operator

bbyreddy:
Update the configuring Edge operator section of the "Edge installation" in a self-managed k8s with the renamed `c8yedge-operator-config`.
As this rename of the ConfigMap is a breaking changes, a release notes is added (though we are almost certain that very few - TSA and may be Edwards use this option currently) for the customers to note when they upgrade Edge to `2026.0.1`.

### DIFF
--- a/content/edge/installing-edge-bundle/install-edge-with-edge-operator.md
+++ b/content/edge/installing-edge-bundle/install-edge-with-edge-operator.md
@@ -64,16 +64,16 @@ To configure proxy settings and trusted certificates, create or update a ConfigM
   - `https_proxy` - HTTPS proxy URL
   - `socks_proxy` - SOCKS proxy URL
   - `no_proxy` - Comma-separated list of domain suffixes, IP addresses, or CIDR ranges that bypass the proxy. This must include:
-      - {{< management-tenant >}} and the Edge tenant domain names
-      - Kubernetes Pod CIDR (Cluster pod IP address range)
-      - Kubernetes Service CIDR (Cluster service IP address range)
-      - Any additional domains, hosts or IP addresses that bypass the proxy
-      Example: 
-        `127.0.0.1,::1,localhost,.svc,.cluster.local,cumulocity,<edge domain names, e.g. management-myown.iot.com,myown.iot.com>,<kubernetes cluster IP range, e.g. 10.43.0.0/16>`
+      - {{< management-tenant >}} and the Edge tenant domain names.
+      - Kubernetes Pod CIDR (Cluster pod IP address range).
+      - Kubernetes Service CIDR (Cluster service IP address range).
+      - Any additional domains, hosts or IP addresses that bypass the proxy.
   - `ca.crt` - One or more trusted TLS certificates in PEM format that the Edge operator and the Edge should trust in addition to publicly known certificate authorities. Multiple certificates can be provided by concatenating them into a single PEM bundle.
 
-#### Example c8yedge-operator-config ConfigMap
+#### Apply changes
+After creating or updating the ConfigMap, restart the Edge operator as described in [Restarting the Edge operator](/edge/manage-edge/#restart-operator)
 
+#### Sample ConfigMap
 ```yaml
 apiVersion: v1
 kind: ConfigMap
@@ -92,6 +92,3 @@ data:
   ca.crt: |
     <CERTIFICATES_TO_TRUST>
 ```
-
-#### Apply changes
-After creating or updating the ConfigMap, restart the Edge operator as described in [Restarting the Edge operator](/edge/manage-edge/#restart-operator)
--- a/content/edge/installing-edge-bundle/install-edge-with-edge-operator.md
+++ b/content/edge/installing-edge-bundle/install-edge-with-edge-operator.md
@@ -56,8 +56,8 @@ For more information about the structure and configuration options available in 
 ### Configuring the Edge operator with a proxy and trusted TLS certificates
 
 You can configure the Edge operator to:
-  - Route outbound traffic through a proxy server when deployed behind a proxy
-  - Trust additional TLS certificates for external endpoints
+  - Route outbound traffic through a proxy server when deployed behind a proxy.
+  - Trust additional TLS certificates for external endpoints.
 
 To configure proxy settings and trusted certificates, create or update a ConfigMap named `c8yedge-operator-config` in the `c8yedge` namespace (or the namespace where Edge is deployed) with the required configuration keys described below:
   - `http_proxy` - HTTP proxy URL
--- a/content/edge/installing-edge-bundle/install-edge-with-edge-operator.md
+++ b/content/edge/installing-edge-bundle/install-edge-with-edge-operator.md
@@ -69,7 +69,7 @@ To configure proxy settings and trusted certificates, create or update a ConfigM
       - Kubernetes Service CIDR (Cluster service IP address range)
       - Any additional domains, hosts or IP addresses that should bypass the proxy
       Example: `127.0.0.1,::1,localhost,.svc,.cluster.local,cumulocity,<edge domain names, e.g. management-myown.iot.com,myown.iot.com>,<kubernetes cluster IP range, e.g. 10.43.0.0/16>`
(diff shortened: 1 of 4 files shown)


content/edge/installing-edge-bundle/install-edge-with-edge-operator.md
```diff
@@ -53,54 +53,64 @@ This command will complete immediately, and the installation will proceed in the
...
+  - `http_proxy` - URL of the HTTP proxy server
+  - `https_proxy` - URL of the HTTPS proxy server
+  - `socks_proxy` - URL of the SOCKS proxy server
+  - `no_proxy` - Comma-separated list of domain suffixes, IP addresses, or CIDR ranges that should bypass the proxy. This list must include:
+      - {{< management-tenant >}} and the Edge tenant domain names
+      - Kubernetes Pod CIDR (Cluster pod IP address range)
+      - Kubernetes Service CIDR (Cluster service IP address range)
+      - Any additional domains, hosts or IP addresses that should bypass the proxy
+      Example: `127.0.0.1,::1,localhost,.svc,.cluster.local,cumulocity,<edge domain names, e.g. management-myown.iot.com,myown.iot.com>,<kubernetes cluster IP range, e.g. 10.43.0.0/16>`
+  - `ca.crt` - One or more certificates in PEM format that the Edge operator and the Edge should trust in addition to the certificates already included in the system trust store. Multiple certificates can be provided by concatenating them into a single PEM bundle.
```
rthrippleton: I'm wondering if using 'certificates' instead of 'certificate authority' is ambiguous here, but I'm not 100% familiar with all of the x509 terminology. Can you take an actual certificate used by a service and put it here, or ca.crt have to be something used to sign?

Also

```suggestion
  - `ca.crt` - One or more certificates in PEM format that the Edge operator and the Edge should trust in addition to publicly known certificate authorities. Multiple certificates can be provided by concatenating them into a single PEM bundle.
```
because the original implies that anything that has been added by hand to the _operating system_ trust store will automatically be trusted by the operator. This is especially risky in the private registry case, because somebody would add the CA for their private registry to the OS trust store, and then assume they don't need to add the same cert for the operator.

content/edge/installing-edge-bundle/install-edge-with-edge-operator.md
```diff
@@ -53,54 +53,64 @@ This command will complete immediately, and the installation will proceed in the
...
-  ca.crt: <CA-CERTIFICATES TO TRUST>
+  # Trusted TLS/SSL certificates in PEM format
+  ca.crt: |
+    <CA_CERTIFICATES_TO_TRUST>
 ```
 
-By configuring Edge with the appropriate proxy settings, you ensure that it can seamlessly communicate with external endpoints through the proxy server, allowing it to function effectively in environments where proxy usage is mandated.
-
-The table below provides more information:
+After creating or updating the ConfigMap, configure the Edge operator to use it by setting the `operatorConfigCMName` Helm value during installation or upgrade:
```
rthrippleton: I'm on the fence about this, but thinking out loud:

I wonder if it's better to put a much less verbose command here, without tediously restating the known values.
e.g.
`helm upgrade ...   --reuse-values --set operatorConfigCMName=...`

That said, I'm not sure if we can avoid `--version`, and that's a problem because `--version=2026` might caused some unnecessary operator bouncing if they're not already on the latest.

Also remove the `login`, because that's already implicit in the fact that they installed it, I think?

content/edge/installing-edge-bundle/install-edge-with-edge-operator.md
bbyreddy: > I'm wondering if using 'certificates' instead of 'certificate authority' is ambiguous here, but I'm not 100% familiar with all of the x509 terminology. Can you take an actual certificate used by a service and put it here, or ca.crt have to be something used to sign?

You can indeed trust either the server/leaf certificate (actual certificate used by the server) or any of the signing certificates (root or intermediary) in the trust chain. `root --> intermediary --> server certificate` 

> because the original implies that anything that has been added by hand to the _operating system_ trust store will automatically be trusted by the operator. This is especially risky in the private registry case, because somebody would add the CA for their private registry to the OS trust store, and then assume they don't need to add the same cert for the operator.

For the private registry to work, they are expected do both. I mean add the server certificate (or one of its chain) to the system trust store (in the node for the `containerd` to trust when it pulls the docker images) and also to the operator (via this ConfigMap) for it to trust while pulling the artifacts (helm charts and ORAS files) from the private registry. 
Adding the same to the JKS truststore passed to the Cumulocity core's JVM, could be our workaround if the core/platform team doesn't agree on providing a solution from their end.

You are right, I changed the text as you suggested --> `in addition to publicly known certificate authorities`  https://github.com/Cumulocity-IoT/c8y-docs/pull/4767/commits/afaa6f160bedacbf53565db21b7f50317943cb93

content/edge/installing-edge-bundle/install-edge-with-edge-operator.md
bbyreddy: Actually I contemplated on moving the section **Configuring the Edge operator with a proxy and trusted TLS/SSL certificates**  before **Installing the Edge operator**. But this also has a caveat, that these configurations can be created or updated post the operator installation. 
In fact, this is the specific reason we had initially decided to keep the ConfigMap name for the operator config as a predefined name. In we keep the name fixed (not configurable), the user just needs to restart the operator after creating/updating the ConfgMap.

bbyreddy: As discussed, the `custom-environment-variables` CM is renamed to `c8yedge-operator-config` and the doc is updated accordingly. The CM name is not configurable and user just has to create or update the CM with a pre-defined name `c8yedge-operator-config`, which the operator reads and updates its configurations accordingly when it is restarted.

content/edge/manage-edge-bundle/registry-password.md
```diff
@@ -8,9 +8,18 @@ sector:
 
 At some point the credentials you use for the registry might change. In that case, Edge needs to be updated so that it can access the artifacts for any future upgrade.
 
-These credentials are stored as a secret within the Kubernetes cluster, and can be modified with `kubectl` commands on the environment in which Edge is installed:
+These credentials are stored as a Kubernetes Secret, and can be updated using `kubectl` command:
```
rthrippleton:
```suggestion
These credentials are stored as a Kubernetes Secret, and can be updated using the `kubectl` command:
```

rthrippleton (approved):

content/edge/installing-edge-bundle/install-edge-with-edge-operator.md
```diff
@@ -53,54 +53,42 @@ This command will complete immediately, and the installation will proceed in the
...
-
-When {{< product-c8y-iot >}} Edge is deployed behind a proxy, it must be configured to communicate with external endpoints over the internet through the proxy server.
-To configure Edge to use a proxy, you must create or update a ConfigMap named `custom-environment-variables` in the c8yedge namespace (or the one you deployed Edge into) with the required proxy settings. The keys `http_proxy`, `https_proxy` and `socks_proxy` must be set to the URLs of the HTTP, HTTPS and Socks proxies, respectively. The key `no_proxy` must be set to specify a comma-separated list of domain suffixes, IP addresses, or CIDR ranges that Edge should bypass the proxy server for.
-
-Here is an example of a ConfigMap with proxy settings:
-
+### Configuring the Edge operator with a proxy and trusted TLS certificates
+
+You can configure the Edge operator to:
+  - Route outbound traffic through a proxy server when deployed behind a proxy
```
BeateRixen:
```suggestion
  - Route outbound traffic through a proxy server when deployed behind a proxy.
```

content/edge/installing-edge-bundle/install-edge-with-edge-operator.md
```diff
@@ -53,54 +53,42 @@ This command will complete immediately, and the installation will proceed in the
...
+  - Trust additional TLS certificates for external endpoints
```
BeateRixen:
```suggestion
  - Trust additional TLS certificates for external endpoints.
```

BeateRixen (approved):

PR #4785, opened 2026-06-01
MTM-53765 Clarify that 1-way TLS is always enabled

scmi-c8y:
Fixing minor user doc issues for the Data in Motion Quality Afternoon:
* [MTM-53765](https://cumulocity.atlassian.net/browse/MTM-53765) Improve MQTT SSL documentation: A customer was confused about whether 1-way SSL was available. Clarified that both 1-way and 2-way SSL are always enabled, and the one used simply depends on whether the client provides a certificate or not.

[MTM-53765]: https://cumulocity.atlassian.net/browse/MTM-53765?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/content/device-integration/mqtt-bundle/implementation.md
+++ b/content/device-integration/mqtt-bundle/implementation.md
@@ -22,7 +22,10 @@ Port 80 is deactivated in cloud systems.
 {{< /c8y-admon-info >}}
 
 Port 8883 supports two types of SSL: two-way SSL using certificates for client authorization and one-way SSL using username and password for client authorization.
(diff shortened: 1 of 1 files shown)


schmiel-cumulocity (approved):
I cannot even count how many times customers, especially NTT were asking about this 👍🏻

BeateRixen (approved):

PR #4793, opened 2026-06-08
[NO-JIRA] ODBC deprecation note; task name naming policy

cheinz-sag:
Hi, can you pls review?


ChrisFurlong:
Can we include the link to the relevant MS article?

cheinz-sag:
> Can we include the link to the relevant MS article?

I was thinking the same, but the MS article https://learn.microsoft.com/en-us/power-query/transition-to-adbc is pretty generic. Better would be a Dremio-specific article on how to move from ODBC to ADBC. Closest one I found is this one: https://www.dremio.com/blog/announcing-arrow-database-connectivity-adbc-in-microsoft-power-bis-connector-for-dremio/
What do you think?

ChrisFurlong:
Think we should include both links - the Dremio one is a good background on ADBC, the MS one has the timescale

BeateRixen (approved):

ChrisFurlong (approved):

PR #4795, opened 2026-06-09
docs: [CDH-6413] Add section  to `using.md` about setting the default…

flsc-c8y:
… base location for the catalog per tenant via tenant option


riemenschneider:
Is this really the correct place for this kind of information? Setting the default base location per environment is documented in the OPS docs. Now we have our of a sudden a small section about the configuration of the default base location per tenant without any context.

eickler:
> Is this really the correct place for this kind of information? Setting the default base location per environment is documented in the OPS docs. Now we have our of a sudden a small section about the configuration of the default base location per tenant without any context.

I was earlier thinking to put it here, but before we have the "bring your own lake" story, there is little that the end user can really do here. So I would park it until then.

flsc-c8y:
> > Is this really the correct place for this kind of information? Setting the default base location per environment is documented in the OPS docs. Now we have our of a sudden a small section about the configuration of the default base location per tenant without any context.
> 
> I was earlier thinking to put it here, but before we have the "bring your own lake" story, there is little that the end user can really do here. So I would park it until then.

Will convert to draft to imply parking the issue.

PR #4870, opened 2026-07-01
Link checker: diagnostics mode, PR trigger, resource mocks, fixes

pawel-rynarzewski-c8y:
## Summary

This branch grew out of investigating a failing "Link Checker" scheduled run
on `develop`.

### Process improvements

**1. Targeted manual runs.** `workflow_dispatch` now accepts `branch` and
`urlsWith` inputs, so you can check a single branch and/or a subset of links
matching a substring, instead of only being able to trigger the full
develop/release matrix. Behavior is unchanged when neither is supplied.

**2. Run the checker from a PR.** Adding the `broken-link-check` label to a
PR now runs the checker against that PR's branch and posts a summary
comment (failing links + link to the full run) directly on the PR. A new
comment is posted per label addition, so the PR history shows when each
check was requested and what it found. This is separate from the scheduled
run's issue-filing flow, which is unaffected.

**3. Built-in diagnostics mode.** Investigating a CI-only failure (loads
fine in a browser, fails only in Actions) used to mean hand-writing
throwaway instrumentation, committing it, dispatching, then reverting.
There's now a `diagnostics` dispatch input that - combined with `branch` +
`urlsWith` - logs every network request's start/completion (so a hang shows
as a "started, never finished") plus console errors/uncaught exceptions,
with zero effect on normal runs. The full Cypress log is now also uploaded
as a build artifact on every run (8-day retention), diagnostic or not.

**4. Mocking known-flaky third-party resources.**
`known-resource-mocks.cjs` is a small declarative registry: pattern +
response + a mandatory `reason` (so the list stays legible), with per-link
overrides for cases where the same domain needs different behavior
depending on which page it's embedded in. Replaces one-off inline
`cy.intercept()` stubs. It's for third-party resources that block
appropriate loading of the pages in Cypress env.

**5. Stop treating third-party console errors as link failures.**
Previously, any uncaught JS exception on a page a link points to failed the
test - the only fix available was appending the exact error message to a
growing allowlist in `cypress/support/e2e.js` (37 entries, purely reactive).
That list is now removed (down to 1 entry, restored after a regression run -
see below). Uncaught exceptions only fail the test when they happen on a
`cumulocity.com` page (the only domain where a JS error might indicate a
real bug we're responsible for, and the only pages the test asserts real
content on). Every other page's console errors are unconditionally
tolerated - link validity is already covered by the test's real assertions
(status code, non-empty body, fragment existence).

**6. README rewritten as a decision guide for the next failure.** Documents
how to tell apart the different failure classes covered above (broken
content, a blocked sub-resource, a target URL that's itself
blocked/rate-limited, a tolerable JS error) and how to confirm which one
you're looking at using diagnostics mode plus cross-referencing sibling
branches/runs, instead of guessing from the error text.

### Bug fixes found along the way

- **Broken anchors in `data-preparation/smart-functions-bundle`**: links used
  `#runtime-behavior-and-limits` (the page title's slug), but the bundle's
  actual anchor comes from the source filename (`limits.md` → `#limits`).
- **`goc8ycli.netlify.app` links timing out in CI only**: root-caused (via a
  live diagnostic run) to `asciinema.org`'s embed script never responding
  from GitHub Actions runners specifically - it still blocks the browser's
  `load` event despite the `async` attribute. Fixed via the new
  `known-resource-mocks.cjs` (item 4 above).
- **`docker.com` links failing in CI**: a Vector marketing tracking pixel
  (`cdn.vector.co/pixel.js`) throws an unhandled rejection when it can't
  establish a visitor ID on a fresh session - unrelated to link validity.
  Now covered by the domain-based exception handling (item 5 above).
- **`cumulocity.com/apama/docs/.../ApamaDoc/` (an old javadoc frameset page)**:
  a full-suite regression run after removing the 37-entry allowlist (item 5)
  found this was the one entry actually needed - `top.loadFrames is not a
  function` is a legacy frame-navigation script error on a `cumulocity.com`
  page, restored as the sole remaining entry.
- **`logback.qos.ch` (4 URLs) failing intermittently, CI-only**: cross-
  referencing sibling branches in the same historical workflow run showed it
  passing fast (2-3s) on two branches while failing all 10 retries
  (consistently spaced at the exact timeout) on a third, at the same time.
  This points to IP-reputation-based blocking on the target site tied to
  whichever ephemeral runner IP GitHub Actions assigns - not a real broken
  link, and not fixable by retrying. Added to `excludedLinks`.

### Housekeeping

- Bumped `actions/checkout` and `actions/setup-node` from `v3` to `v4` to address
  the "Node.js 20 is deprecated" warning surfaced on every run.

## Test plan

- [x] All of the above verified locally and via live GitHub Actions dispatch
      runs (targeted `branch`/`urlsWith`/`diagnostics` combinations) during
      development - see commit messages for specifics on each.
- [x] Full unfiltered run on this branch (no `urlsWith` filter): confirmed
      clean at `1413 tests, 1398 passing, 0 failing, 15 skipped` (11
      pre-existing exclusions + 4 new `logback.qos.ch` exclusions) - no
      unexpected failures, and no case where removing an allowlist entry
      broke something else.


broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -1,11 +1,30 @@
 const allUrls = require('../../all_links.json');
+const KNOWN_RESOURCE_MOCKS = require('../../known-resource-mocks.cjs');
+
+// Domains we actually own/publish - a JS error on one of these could mean a
+// real bug worth catching. Everywhere else, an uncaught exception isn't
+// signal about whether the link is valid (see the describe-level
+// uncaught:exception handler below); what matters there is already covered
+// by the test's real assertions (status code, non-empty body, fragment
+// existence).
+const OWN_DOMAINS = ['cumulocity.com'];
```
jdre-c8y: just a suggestion- should we make separate file for own domains too? not sure, if it's only one and we don't plan to get more, maybe not...

broken-links-script/cypress/e2e/link-checker.cy.js
jdre-c8y: same as above- should excluded links be in separate file? same concern, same doubts, but raising idea.

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -68,6 +98,79 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
...
+    if (!DIAGNOSTICS) return;
+    cy.intercept('**', (req) => {
+      try {
+        const start = Date.now();
+        diagnosticLog.push(`START ${req.method} ${req.url}`);
+        req.on('response', (res) => {
+          try {
+            diagnosticLog.push(`DONE ${res.statusCode} ${Date.now() - start}ms ${req.url}`);
+          } catch (e) {
+            // never let logging break the request
```
jdre-c8y: still, shouldn't we push error to diagnostics?

broken-links-script/cypress/e2e/link-checker.cy.js
```diff
@@ -68,6 +98,79 @@ describe('Link and Routing Validation - Individual URL Checks', () => {
...
+    }
+    const onBeforeLoad = (win) => {
+      options.onBeforeLoad?.(win);
+      ['error', 'warn'].forEach((level) => {
+        const original = win.console[level];
+        win.console[level] = (...args) => {
+          try {
+            diagnosticLog.push(`console.${level}: ${args.map(String).join(' ')}`);
+          } catch (e) {
+            // ignore
```
jdre-c8y: same as above

broken-links-script/README.md
```diff
@@ -192,33 +198,88 @@ You can also leave a short reason above it:
 
 Use this for links that should be skipped completely.
 
-### 2. Add known browser exceptions
+### 2. Mock a known-problematic third-party resource
 
-Some third-party sites throw JavaScript errors that do not actually mean the page is broken.
+Sometimes the link under test is fine, but the page it points to embeds a
+third-party resource (analytics, an embed widget, a tracking pixel, etc.)
```
jdre-c8y: embed or embedded?

broken-links-script/README.md
```diff
@@ -192,33 +198,88 @@ You can also leave a short reason above it:
...
+Uncaught JS exceptions are already tolerated automatically for any page
+outside `cumulocity.com` (see `OWN_DOMAINS` in `cypress/e2e/link-checker.cy.js`)
+- a third-party site's own console errors (analytics pixels, chat widgets,
+etc.) aren't signal about whether the link pointing to it is valid, so we
+don't fail on them regardless of message text. What still matters for those
+pages is covered by the test's real assertions: status code, non-empty body,
+fragment existence.
+
+`cypress/support/e2e.js` is now only for the rare case of a genuine
+**`cumulocity.com`-domain** exception that's confirmed harmless. Before
```
jdre-c8y, line 264: related to my previous comment- to align if own domains will get own file

broken-links-script/cypress/e2e/link-checker.cy.js
pawel-rynarzewski-c8y: yes, pushing it there

broken-links-script/cypress/e2e/link-checker.cy.js
pawel-rynarzewski-c8y: yes, pushing it there

broken-links-script/README.md
pawel-rynarzewski-c8y: corrected

broken-links-script/cypress/e2e/link-checker.cy.js
pawel-rynarzewski-c8y: Separate file just with the list of our domains? Or do you mean a file per domain with the list of known exceptions on this domain? Either way, honestly, I'd like to avoid that. I want to keep these lists of exceptions as minimal as possible, don't want them to grow like the one in previous approach.

If the main concern is mixing checker logic and config options, maybe this would solve it - 8c8a40a76?

jdre-c8y: okay, I'm fine with it

jdre-c8y (approved):

BeateRixen (approved):

pawel-rynarzewski-c8y:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|release/y2026|[](https://github.com/Cumulocity-IoT/c8y-docs/pull/4935)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport)

pawel-rynarzewski-c8y:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|release/y2025|[](https://github.com/Cumulocity-IoT/c8y-docs/pull/4936)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport)

PR #5013, opened 2026-08-07
CIE-5643 - Caution around changing the IP address

rthrippleton:


BeateRixen (approved):

content/edge/manage-edge-bundle/changing-ip.md
```diff
@@ -0,0 +1,15 @@
...
+  - edge
+---
+
+{{< c8y-admon-caution >}}
+This section applies only to an installation managed by the **c8yedge** tool. If you install Edge on a self-managed Kubernetes cluster, host network configuration is outside the scope of this documentation.
+{{< /c8y-admon-caution >}}
+
+Edge runs on top of the Kubernetes platform. Kubernetes is server infrastructure and not adapted for dynamic network environments, unlike desktop applications. Changing the IP address of the host that Edge is running on will lead to undefined behavior.
+
+To avoid this, immediately reboot the host after changing the IP address.
```
raghuvamsi-d, line 15: Can we suggest them if they are using hypervisor then to keep the IP static  like to avoid bridging of they expect network to change from time to time?

rthrippleton: I've pushed something quite vague and generic. Can you suggest some technical wording that might direct somebody to the write part of their hypervisor configuration? I'm not very familiar with setups like this.

raghuvamsi-d: This sounds correct to me, a small word of caution will make it further clearer.

Note: This approach is applicable when the VM uses NAT or an internal/private virtual network provided by the hypervisor. When the virtual network interface is configured in bridged mode, the VM shares the host's physical network and subnet. In this case, changes to the host's network configuration or connectivity can affect the VM, and a statically configured guest IP address may become unreachable or cause connectivity issues.

rthrippleton: How about bca6af1 ?

raghuvamsi-d: This looks accurate to me. I am approving this PR.

raghuvamsi-d (approved):

PR #5022, opened 2026-08-10
Update proxy configuration documentation for Go and JVM

raghuvamsi-d:


rthrippleton (commented):
I'm hoping that we can automate all of this as part of the ticket, so then this becomes unnecessary.

But if we can't automate all of this, I'd suggest at least automating the CIDR vs. wildcard aspect. Having to have the user double-up on configuration just to fit our internal implementation details is very annoying.

So somebody writes `10.42.0.0/16' in the config, we should implicitly be generating the `10.42.*` ourselves.

content/edge/installing-edge-bundle/install-edge-with-edge-operator.md
```diff
@@ -85,8 +87,10 @@ data:
   https_proxy: <HTTPS Proxy URL>
   socks_proxy: <SOCKS Proxy URL>
 
-  # Comma-separated list of domain suffixes, IP addresses, or CIDR ranges that should bypass the proxy
-  no_proxy: 127.0.0.1,::1,localhost,.svc,.cluster.local,cumulocity,<edge domain names, e.g. management-myown.iot.com,myown.iot.com>,<kubernetes cluster IP range, e.g. 10.43.0.0/16>
+  # Comma-separated list of domain suffixes, IP addresses, or CIDR ranges that should bypass the proxy.
+  # Specify each Pod and Service CIDR in both formats: subnet/CIDR notation (for example, 10.42.0.0/16)
+  # for Go-based components, and wildcard notation (for example, 10.42.*) for JVM-based components.
+  no_proxy: 127.0.0.1,::1,localhost,.svc,.cluster.local,cumulocity,<edge domain names, e.g. management-myown.iot.com,myown.iot.com>,<kubernetes cluster IP ranges, e.g. 10.42.0.0/16,10.42.*,10.43.0.0/16,10.43.*>
```
rthrippleton, line 93: Two `e.g.` in a row doesn't scan right here. Put the whole example at the end.

raghuvamsi-d:
> I'm hoping that we can automate all of this as part of the ticket, so then this becomes unnecessary.
> 
> But if we can't automate all of this, I'd suggest at least automating the CIDR vs. wildcard aspect. Having to have the user double-up on configuration just to fit our internal implementation details is very annoying.
> 
> So somebody writes `10.42.0.0/16' in the config, we should implicitly be generating the `10.42.*` ourselves.

@rthrippleton I got a hunch on how to do it. I am trying out the changes for the same. Hopefully we may not even needed to publish this if it works.

content/edge/installing-edge-bundle/install-edge-with-edge-operator.md
BeateRixen, line 93: Also, please use "for example" instead of "e.g.".

PR #5097, opened 2026-08-31
MTM-67491: Document SmartREST MQTT version 5.0 behaviour

schmiel-cumulocity:
## Summary
- MQTT Service supports MQTT 3.1.1 and MQTT 5.0 for SmartREST/Core MQTT topics
- Documents that Core MQTT topic handling (SmartREST, JSON via MQTT) is identical for both protocol versions: errors reported only via `s/e` or `error` topics (not MQTT 5.0 reason codes), disconnect behavior unchanged, MQTT 5.0 publish properties not used/forwarded

## Test plan
- [ ] Docs build renders new section correctly
- [ ] Anchor links (`#core-mqtt-mqtt-5-behavior`, `#mqtt-error-reporting`, `#mqtt-50-publish-features`) resolve


BeateRixen (approved):

scmi-c8y (approved):
Praise: A very clear statement of the limitations. Probably nobody will read it before they get themselves into trouble, but we have it covered when they do ;-)

PR #5099, opened 2026-09-01
CTM-2656: DTM Data Service SLA and SLI

thomaswinkler:


content/service-terms/service-level-bundle/dtm-data-service-sla.md
```diff
@@ -0,0 +1,100 @@
...
+  × average number of linked assets matched per device measurement
+```
+
+The multiplier between the two rates is called **fan-out**. Fan-out is entirely determined by Customer configuration and varies by orders of magnitude between tenants or even assets depending on their configured linked data points: a device measurement carrying more series can match more linked data point, and each linked asset adds another created measurement. A device sending 10 measurements per second whose data points are linked to 30 assets produces an upstream measurement rate of 300 per second, not 10.
+
+Because fan-out is Customer-defined, the service-level objectives below apply to the upstream measurement rate, not to the device measurement rate.
+
+### Customer responsibilities
+
+To ensure the successful operation of the Service, you must fulfill the following responsibilities.
```
BeateRixen, line 39: The whole paragraph switches to "you" instead of "Customer". This is a bit confusing and also inconsistent with how we handle it in other articles.

content/service-terms/service-level-bundle/dtm-data-service-sla.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Service description
+
+The Service propagates measurements from devices to the assets they are linked to in the Digital Twin Manager asset hierarchy, so that operational data can be consumed in asset context by dashboards, analytics, smart rules, and business applications. The Service performs the following core functions:
+
+* Processes all incoming device measurements of the tenant in real time.
+* Resolves the [linked data points](/dtm/asset-hierarchy/#datapoints) configured in the Digital Twin Manager application for the measurement source.
+* Creates one asset-side measurement per linked target asset, carrying the value, unit, type, and time of the source measurement under the target fragment and series defined by the linked data point.
+
+The Service is a stateless, best-effort forwarding pipeline. It does not store measurements. The device measurement and the created asset measurement are both stored by the {{< product-c8y-iot >}} platform and are subject to the [Platform service-level agreement](/service-terms/service-level/#platform-sla) and to the retention rules configured by Customer.
```
eickler, line 21: [q] What does the best effort refer to here?

content/service-terms/service-level-bundle/dtm-data-service-sla.md
```diff
@@ -0,0 +1,100 @@
...
+
+The load the Service puts on the platform is not determined by the number of device measurements Customer sends, but by the number of measurements the Service creates from them. This number is called the **upstream measurement rate** and is the single quantity against which the Service is dimensioned and measured:
+
+```
+upstream measurements per second
+  = device measurements per second
+  × average number of linked assets matched per device measurement
+```
+
+The multiplier between the two rates is called **fan-out**. Fan-out is entirely determined by Customer configuration and varies by orders of magnitude between tenants or even assets depending on their configured linked data points: a device measurement carrying more series can match more linked data point, and each linked asset adds another created measurement. A device sending 10 measurements per second whose data points are linked to 30 assets produces an upstream measurement rate of 300 per second, not 10.
```
eickler, line 33: [m] This paragraph is broken (grammar and content) and/or not understandable

content/service-terms/service-level-bundle/dtm-data-service-sla.md
```diff
@@ -0,0 +1,100 @@
...
+
+To ensure the successful operation of the Service, you must fulfill the following responsibilities.
+
+#### Load management
+
+You are responsible for the upstream measurement rate your tenant generates, and therefore for all three factors that produce it:
+
+* **Ingestion rate**: The number of measurements your devices send per second.
+* **Measurement shape**: The number of fragments and series carried by each measurement. Every series that matches a linked data point contributes to fan-out.
+* **Linking configuration**: The number of assets each device data point is linked to. Every additional link on a data point multiplies the load produced by every measurement carrying that data point.
```
eickler, line 47: I think no customer questions their responsibility for the ingestion. I think the two sections can be simplified.
How about sticking with the existing structure in the other docs and calling the section "customer responsibilities"?

eickler (commented):
dropping the comments that were still pending

content/service-terms/service-level-bundle/dtm-data-service-sla.md
```diff
@@ -0,0 +1,100 @@
...
+
+#### Service-level objectives
+
+The following objectives measure the quality of the Service:
+
+| Service level indicator  | Monthly target                                |
+| ------------------------ | --------------------------------------------- |
+| Service availability     | As per [Platform service-level agreement](/service-terms/service-level/#service-availability) |
+| Propagation latency      | 95 percentile of sustained load ≤ 30 seconds  |
+| Propagation completeness | ≥ 99.9% of sustained load                     |
```
eickler, line 86: Call it durability?

content/service-terms/service-level-bundle/dtm-data-service-sla.md
thomaswinkler: @BeateRixen What other articles are you referring to? I was using other SLA documens as reference and there it is exactly the same. See for example: https://cumulocity.com/docs/service-terms/service-level/#streaming-lake-ingestion-sla

I can change it to customer, but than it would be inconsistent.

BeateRixen: Sorry, I meant paragraphs in this document, not articles. It might be wrong in other documents as well :-). @eickler What is your take? Is it ok for you to use "you" in this section instead of "Customer"?

PR #1324, opened 2023-08-18
Dtm links

MWindel:
* Checked and fixed links, file names and anchors

PR #1323, opened 2023-08-18
Steaming analytics links

dianaolives:

PR #5143, opened 2026-09-11
docs: [CDH-7052] document Streaming Lake Ingestion with your own data lake

riemenschneider:
## Why

Streaming Lake Ingestion can write into a customer's own S3 bucket or ADLS Gen2 storage account, and none of it was documented for customers — the bundle did not mention the option at all, and two pages actively said the service has no configuration. Provisioning also changed on both clouds: AWS is new end to end (the customer creates a role, its trust policy names the environment's base principal, and an External ID has to match that policy character for character), and Azure gained directory ID capture and admin consent. Until now this was covered only by internal runbooks and a PDF handed to one customer.

## Changes

- A new **Using your own data lake** page carries both procedures: on AWS the bucket, the role with its two policies and the External ID; on Azure the storage account requirements, the admin consent, the two role assignments at their two different scopes, and soft delete. The base-location rules, what the setup verifies, troubleshooting, adding further tenants and revoking access are shared between the two clouds rather than written twice, which is why this is one page and not one per cloud — say so in review if the navigation would be better served by a split.
- The four existing bundle pages gain pointers to it, and the two statements that said the service has no configuration now distinguish *what* it offloads (not configurable) from *where* it writes (this page).
- Resolution tiers, environment variable names and REST endpoints are deliberately absent; the customer does all of this in **Settings** > **Data Lake**.

## Rebased and corrected against the landed implementation

The first draft was written ahead of the implementation by agreement. CDH-6980, CDH-6932 and CDH-7053 have since merged and both customer runbooks were rewritten alongside them, so the second commit reconciles the page with what actually shipped. CDH-7039 is the one exception: [Cumulocity-IoT/iceflow#1366](https://github.com/Cumulocity-IoT/iceflow/pull/1366) makes provisioning reject a role name without the `c8y-streaming-lake-ingestion-` prefix, and it is approved rather than merged. The page states the prefix as a requirement regardless, because it is the base principal's IAM policy that enforces it, not the validator. What changed for the customer:

- **Azure verifies less than the page claimed.** For a tenant that brings its own Entra directory, the setup does not check the account's region, soft delete, or hierarchical namespace — reading those needs **Reader** on the storage account, which this onboarding deliberately does not ask for. The page had said provisioning refuses a flat account and that a region mismatch fails the setup; both are gone. The account requirements moved into the prerequisites with an `isHnsEnabled` self-check, and an admonition now states plainly which three properties are unverified and why.
- **The AWS region *is* enforced** (every S3 request is signed for the deployment's region), so that asymmetry between the clouds is now stated rather than smoothed over.
- **The AWS role must also sit at the default path.** A role created under a path cannot be assumed at all, because the path occupies the position the required `c8y-streaming-lake-ingestion-` prefix has to be in. Roles cannot be renamed either, so both mistakes mean delete and recreate.
- **Azure storage account requirements added**: `StorageV2`, hierarchical namespace, public network access, anonymous access off — plus a caution that locking the account to selected networks needs the environment's egress identity from support first, because doing it unilaterally stops ingestion with an error that names nothing about the firewall.
- **The setup page needs `ROLE_OFFLOADING_ADMIN` or `ROLE_TENANT_ADMIN`**, not tenant admin alone, worded to match the Manager API section that landed on `develop` in the meantime.
- Smaller: the wizard takes bucket and prefix as separate fields; `aws s3 mb` replaces `aws s3api create-bucket`, whose `LocationConstraint` flag `us-east-1` rejects; the AWS console path now names the two traps in the role form (**Custom trust policy** rather than *AWS service*/*AWS account*, and attaching no policy until the role exists).

CDH-6696 is also fixed, so the wizard's AWS policies are no longer stale and the page no longer has to route around them.

## Still worth a reviewer's eye

- The wizard now **renders these runbook sections itself**, so this page and the in-product guide are two renderings of the same procedure and can drift. The page says so and positions itself as the version to read before starting or away from the browser.
- Nothing claims that changing the External ID is refused: no change rule exists for it yet (Cumulocity-IoT/iceflow#1287 found this).
- The data-protection asymmetry is deliberate: bucket versioning stays a recommendation on AWS while soft delete is stated as required for production on Azure.

## Screenshots

Five are needed and none can be taken yet — the setup page is not final. Each spot is marked with an HTML comment naming the intended path and caption, so nothing renders as a broken image in the meantime. They want a follow-up pass once the UI is done.

## Verification

Hugo is not available locally, so the bundle was checked with a script for front matter, unknown or unbalanced shortcodes, duplicate and dangling anchors, heading levels (the bundle layout generates `h1`/`h2`, so page content starts at `h3`), and ragged table rows. The only findings are four pre-existing ragged rows in `using.md`'s example query output, confirmed identical on `develop`.


content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,520 @@
+---
+weight: 15
+title: Using your own data lake
+layout: redirect
+---
+
+By default, Streaming Lake Ingestion writes your IoT data into a data lake that {{< company-c8y >}} operates for you. Alternatively, you can have it write into object storage of your own: an **AWS S3 bucket** or an **Azure Data Lake Storage Gen2 storage account** that belongs to your organization.
+
+The data then lives in your account. You create the access grant, you scope it, and you remove it when you want the data flow to stop. This section describes what to set up in your cloud account and how to complete the setup in {{< product-c8y-iot >}}.
```
riemenschneider: That's not really how we designed the process. If you remove your own bucket were SLI is writing to ingestion will stop (break). But that is a failure state. Stopping the data flow is done via unsubscribing the Offloading microservice.

But I just saw the according caution block in #own-lake-removing-access. Therefore it might be OK.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,520 @@
...
+
+Which of the two clouds applies to you is not a choice you make here: it follows the {{< product-c8y-iot >}} environment your tenant runs in. The setup page named below states which cloud it expects, and provisioning refuses storage belonging to the other one.
+
+{{< c8y-admon-info >}}
+You perform the setup yourself, in the **Administration** application under **Settings** > **Data Lake**. Your {{< product-c8y-iot >}} user must have the ROLE_OFFLOADING_ADMIN or the ROLE_TENANT_ADMIN permission. The **OFFLOADING_ADMINISTRATOR** global role carries ROLE_OFFLOADING_ADMIN: assign it to a user in the Administration application under **Accounts** > **Roles**.
+
+The page is also where the values specific to your environment appear — the identity your grant has to name, the region the environment runs in, and, on AWS, a suggested External ID. Those values differ per environment, so this documentation refers to them rather than repeating them.
+
+The page carries the same procedure as a setup guide alongside the fields, with your own values substituted into the commands. This section is the version to read before you start, or away from the browser.
```
riemenschneider: I know I am not a native speaker but the last sentence sounds strange.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,520 @@
...
+
+* **Everything {{< company-c8y >}} can reach, you granted.** No access exists in your account until you create it, and nothing else in your account or directory is touched.
+* **Nothing long-lived is handed over.** Every read and write runs on temporary credentials derived from your grant. You never give {{< company-c8y >}} an access key or a storage account key, and no such field exists.
+* **The grant is per environment.** The identity named in your grant belongs to one {{< product-c8y-iot >}} environment, so a grant made for one environment cannot be used by another.
+
+### Terminology {#own-lake-terminology}
+
+|Term|What it means here|
+|:---|:---|
+|**Base location**|Where your data is written. On AWS `s3://<bucket>/<prefix>`, on Azure `abfss://<container>@<account>.dfs.core.windows.net/<path>`. {{< product-c8y-iot >}} appends the tenant's ID, so that tenant's tables live under `<base location>/<tenant-id>/`|
```
riemenschneider: Should we show in any way that prefix and path also can be empty? (SLI will write to bucket/container root.)

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,520 @@
...
+#### On Microsoft Azure {#own-lake-azure-prerequisites}
+
+|You need|Why|
+|:---|:---|
+|**Cloud Application Administrator** or **Application Administrator** in the storage Entra directory (**Privileged Role Administrator** also works)|To approve the application|
+|**Owner** or **User Access Administrator** on the storage account, or any role that can write role assignments|To assign the two storage roles|
+|Permission to create a storage account and container in the subscription (**Contributor**, or any role that can write storage accounts)|To create the account as described below|
+|Certainty about which Entra directory that storage account's subscription belongs to|Every step happens in that directory, and its ID is one of the values you enter|
+
+The storage account itself has to meet four requirements, and the first of them cannot be corrected afterwards:
```
riemenschneider: In the table the first two requirements are mark as cannot be corrected afterwards.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,520 @@
...
+|**Account kind `StorageV2`**|Hierarchical namespace requires it|No|
+|**Public network access enabled**|{{< company-c8y >}} connects from outside your virtual network, so the account's endpoints have to be reachable from it|Yes — but see the caution below|
+|**Anonymous access disabled**|A different setting despite the similar name: it decides whether anyone with the URL can read your blobs **without credentials**. {{< company-c8y >}} never does — it authenticates as the application you approve and reads with the roles you grant|Yes|
+
+Redundancy and performance tier are yours to choose; nothing here depends on either. Neither does the region — unlike on AWS, provisioning does not refuse an account in another region. Keep it in the same region as the {{< product-c8y-iot >}} environment anyway, or you pay cross-region egress on every read a query engine makes.
```
riemenschneider, line 81: That sounds like in Azure environment and stroage account in the same region are not enforced (only in AWS). Is this really correct? Check CDH-6980 and according PRs.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,520 @@
...
+
+### What the setup verifies {#own-lake-verification}
+
+The setup does not take your word for the grant. It exercises it, end to end, through the identity that will actually be used — rather than inspecting your cloud account's settings, which this onboarding gives it no permission to read:
+
+1. **The grant.** On AWS it assumes your role with your External ID; on Azure it acquires a token for your directory as the consented application.
+2. **The data path.** With those credentials it lists, writes, reads back and deletes a test object under the base location.
+3. **The catalog.** Your tenant's Iceberg catalog is created against the values you entered.
+4. **Credential vending.** It then checks that the catalog can hand out the short-lived credentials queries need — a scoped session on AWS, a user-delegation token on Azure. This step comes last because credentials are vended per table, so there has to be a table to ask about: the check creates a temporary namespace and table named `iceflow_provisioning_check`, requests credentials for it, and removes both afterwards, whether the check passed or failed. If you audit your storage, that is what those short-lived objects were.
```
riemenschneider: Is this still correct after #1323 being merged?

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
riemenschneider: Rewritten. Stopping the data flow is now stated as unsubscribing the tenant; withdrawing the grant or removing the storage is called out as a failure state rather than an off switch, with a pointer to Removing access.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
riemenschneider: Reworded: "The page also carries this procedure as a setup guide next to the fields, with your own values already substituted into the commands. Use that while you work through the setup; use this section to read ahead, or when you are not in front of the page."

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
riemenschneider: Added to the Base location row: the prefix or path may be left empty, in which case data is written at the root of the bucket or container. The tenant ID is appended either way.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
riemenschneider: Fixed - now reads "the first two are set when the account is created and cannot be corrected afterwards".

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
riemenschneider: Checked again against merged main, and it is correct — no change made.

`AzureValidator.checkPermissions` branches on the directory: the ARM lookups (region, soft delete, hierarchical namespace) run only when `directory.equalsIgnoreCase(configured.getTenantId())`, i.e. when the storage account sits in *our* Entra directory. For a BYODL tenant, which by definition brings its own, the else branch logs "Storage account properties not checked: reading them needs Reader in the customer's directory, which the onboarding contract does not grant" and skips all three. What still runs for every tenant is the data-plane probe — `checks.blobs().check(container, prefix)` and `checkUserDelegationKey()`.

The reason is the grant we ask for: Storage Blob Data Contributor on the container plus Storage Blob Delegator on the account, no Reader. ARM's effective-permissions query would fail for a *correctly* configured tenant.

#1324 does add a `HIERARCHICAL_NAMESPACE_DISABLED` failure code, which looks like it contradicts this, but it is raised from `AzureStorageAccountPermissionLookupImpl` — the ARM lookup — so it only ever fires on the own-directory branch.

CDH-6980's acceptance criteria do say an Azure account in another region should fail; that is the one part of section 3 the implementation could not deliver for BYODL tenants. AWS is genuinely different: SigV4 binds each request to the deployment's region, so the permission probe reports a non-co-located bucket by itself, with no separate lookup.

Happy to file a ticket for the gap if you want the AC tracked rather than closed as-implemented.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
riemenschneider: Good catch — it was wrong, and is now corrected.

I wrote that paragraph from #1323's description ("creates a throwaway namespace and table, loads it with the delegation header, and drops both with purge"), but the merged code deliberately reversed it. `VendedCredentialsVerification` now documents the table as "One empty table per tenant, created once and left in place": `DROP_TABLE_WITH_PURGE` needs `TABLE_WRITE_DATA`, which neither the management principal nor the tenant's read-only role holds, and an unpurged drop would abandon metadata in the customer's storage on every call where no orphan-file pass would reach it. A 409 on re-run is the steady state.

The page now says the check creates an empty `vending_check` table in an `iceflow_provisioning_check` namespace and leaves it there, why it is not dropped, and that a customer who finds it while auditing their storage should leave it alone. The short-lived object that *is* cleaned up is the `vendingcheck<uuid>` principal, which I have not mentioned — it never appears in their storage.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
riemenschneider: What do you mean with "... use this section to read ahead, or when you are not in front of the page."? I still don't get it.

riemenschneider: Fair — the sentence was trying to tell you when to use the in-product guide versus this page, which is not a distinction a customer needs drawn for them. I have cut it. The note now ends:

"The page shows the steps below as a setup guide next to the fields, with those values already filled into the commands."

That is the whole point I was reaching for: the same steps appear in the product, pre-filled.

BeateRixen:
@riemenschneider Please let me know if I should do a quick review. I usually don´t touch PRs while they are in draft mode.

content/datahub/streaming-lake-ingestion-bundle/overview.md
```diff
@@ -8,7 +8,7 @@ layout: redirect
...
+The Streaming Lake Ingestion feature in DataHub bridges the gap between real-time IoT data and large-scale business analytics and artificial intelligence needs. It takes the continuous stream of IoT data flowing into {{< product-c8y-iot >}} and automatically transforms it into a queryable and AI-ready format in your data lake. This fully managed process requires no configuration to start. When enabled, it processes your data transparently, continuously, reliably, and scalably, making it available for analysis within minutes. If you would rather keep the data in object storage of your own, that is a one-time setup in your own cloud account, described in [Using your own data lake](#own-data-lake).
```
eickler: [m] We currently do not have "managed lake" as option, but this text sounds like it's the customer's choice. I would rephrase the section that it requires no configuration beyond pointing to a folder on your object store as a one-time setup.

content/datahub/streaming-lake-ingestion-bundle/using.md
```diff
@@ -9,6 +9,8 @@ Streaming Lake Ingestion is an optional service in {{< product-c8y-iot >}}. To s
 * Your current device and asset inventory is downloaded from the operational store into the lake.
 * All new incoming data is stored in the lake.
 
+By default, the data is written into a data lake that {{< company-c8y >}} operates for you. To have it written into object storage of your own instead, see [Using your own data lake](#own-data-lake).
```
eickler: [M] Remove -- see other comments.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
+---
+weight: 15
+title: Using your own data lake
+layout: redirect
+---
+
+By default, Streaming Lake Ingestion writes your IoT data into a data lake that {{< company-c8y >}} operates for you. Alternatively, you can have it write into object storage of your own: an **AWS S3 bucket** or an **Azure Data Lake Storage Gen2 storage account** that belongs to your organization.
```
eickler: [M] See general comment -- default is your own lake.
[m] The terms in use here are not consistent, either use "AWS S3 bucket" and "Azure Data Lake Storage Gen2 container", or "AWS S3" and "Azure Data Lake Storage Gen2". A storage account is not an object storage.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+The data then lives in your account, and the access {{< company-c8y >}} has to it is a grant you create, scope and can withdraw. This section describes what to set up in your cloud account and how to complete the setup in {{< product-c8y-iot >}}.
+
+To stop the data flow, unsubscribe the tenant from Streaming Lake Ingestion. Withdrawing the grant or removing the storage instead leaves the service writing into something it can no longer reach, which is a failure state rather than an off switch — see [Removing access](#own-lake-removing-access).
```
eickler: [m] I would phrase it different: "If you only remove the grant or the storage, the service treats this as a temporary failure state and will still apply the service surcharge for Streaming Lake Ingestion."

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+Which of the two clouds applies to you is not a choice you make here: it follows the {{< product-c8y-iot >}} environment your tenant runs in. The setup page named below states which cloud it expects, and provisioning refuses storage belonging to the other one.
```
eickler: [m] That sounds rather impolite in a customer documentation. I would rather write something along the lines of "The service requires you to use the same cloud and cloud region in which your tenant runs in and guides you accordingly during the setup."

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+{{< c8y-admon-info >}}
+You perform the setup yourself, in the **Administration** application under **Settings** > **Data Lake**. Your {{< product-c8y-iot >}} user must have the ROLE_OFFLOADING_ADMIN or the ROLE_TENANT_ADMIN permission. The **OFFLOADING_ADMINISTRATOR** global role carries ROLE_OFFLOADING_ADMIN: assign it to a user in the Administration application under **Accounts** > **Roles**.
```
eickler: [m] Remove the ROLE_TENANT_ADMIN. It's nowhere documented.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+The page is also where the values specific to your environment appear — the identity your grant has to name, the region the environment runs in, and, on AWS, a suggested External ID. Those values differ per environment, so this documentation refers to them rather than repeating them.
```
eickler: [m] Remove this sentence. It's not on the same page (but in the same app) and also the concepts are not yet mentioned.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+The page shows the steps below as a setup guide next to the fields, with those values already filled into the commands.
```
eickler: [m] Remove the sentence.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+{{< /c8y-admon-info >}}
+
+### What you set up, and what it means {#own-lake-what-you-set-up}
+
+On both clouds the shape is the same. One identity belonging to {{< company-c8y >}} is allowed to write into one location in your storage, and everything it can reach is something you granted.
```
eickler: [q] To my knowledge, entra ID admins and lake users are different people in most companies. Should we take that into account in the description?

eickler (commented):
General comments: 

[M] I would structurally have a running section for AWS and a running section for Azure, on different pages. (Or as tabs switching between the steps and terminology on the same page like many do nowadays? Can we do that in our markdown variant?) This is very complex material to digest for the readers, and I would rather not mix AWS and Azure on the same page. A reader is normally only interested in one or the other.

[M] We do not have a managed lake offering in the public preview. It's not a default. This should be correctly reflected in the document, i.e., by having a section "Enabling streaming lake ingestion" followed by "Using streaming lake ingestion". (Instead of showing "Using your own data lake" as an option.) We might need to edit it later, but we have no decision on that topic yet.

[M] Please review the entire text for brevity, in particular for redundancy with the description texts in the user interface. This documentation can have explanations and clarifications, but it does not need to repeat what is in the user interface in all detail.

[m] If we want to provide a view for customers not using our user interface but provisioning tenants themselves automatically, it might also make sense to keep these on separate pages. 

Note: I did not review the technical steps. I appreciate if there's another review.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+* **What is refused is one tenant nested inside another's.** If the location your tenant would write to contains, or is contained by, one another tenant on the environment already holds, the setup refuses it and asks you to pick another — nothing is created, and your storage is not touched. For privacy the refusal does not name the other tenant, since it may belong to a different customer. Choose a location that neither contains nor sits inside the other one.
+* **The base location is fixed once the setup has succeeded.** It is recorded when your tenant's Iceberg catalog is created, and running the setup again does not rewrite it. Entering a different one later therefore moves nothing: the tables stay where they are, and anything already written under the old path would be stranded rather than migrated. [Contact {{< company-c8y >}} support](/additional-resources/contacting-support/) to change it. The same holds for the storage type and, on Azure, the Entra directory.
+
+### Changing the role or the External ID later {#own-lake-rotating-the-grant}
```
eickler: [q] Is that required functionality to have? I noticed that it shows up in the UI now.

content/datahub/streaming-lake-ingestion-bundle/faq.md
```diff
@@ -6,7 +6,9 @@ layout: redirect
 
 ### How can I configure Streaming Lake Ingestion? {#configure-streaming-lake-ingestion}
 
-Streaming Lake Ingestion has no configuration options and stores all incoming realtime data into your lake. Depending on your use case, you can use a combination of options outside of Streaming Lake Ingestion to influence how data appears in your lake:
+Streaming Lake Ingestion stores all incoming realtime data into your lake and has no configuration options for *what* it offloads or *how* the data is shaped. The one thing you can configure is *where* it writes: by default a data lake that {{< company-c8y >}} operates, or object storage of your own, as described in [Using your own data lake](#own-data-lake).
```
y5008720-c8y: is there a C8Y-owned lake option already?

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+Redundancy and performance tier are yours to choose; nothing here depends on either. Neither does the region — unlike on AWS, provisioning does not refuse an account in another region. Keep it in the same region as the {{< product-c8y-iot >}} environment anyway, or you pay cross-region egress on every read a query engine makes.
+
+{{< c8y-admon-caution >}}
+**Locking the account down to selected networks is possible, but not by you alone.** {{< company-c8y >}} reaches your account from its own network, so a firewall that allows only your virtual networks shuts ingestion out. Permitting it means adding the environment's egress identity to your network rules, and that value differs per environment and can only come from {{< company-c8y >}}. [Contact support](/additional-resources/contacting-support/) before restricting network access — do it first, and provisioning or a running tenant stops with a connectivity error that names nothing about the firewall.
```
y5008720-c8y: cumbersome wording: 
- remove the part " — do it first, and provisioning or a running tenant stops with a connectivity error that names nothing about the firewall."
- change "is possible, but not by you alone" to something like "is only possible in collaboration with C8Y support"

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+     Caption: "The Data Lake setup page under Settings in the Administration application"
+     Blocked: the setup page is not final yet. -->
+
+### Setting up your AWS account {#own-lake-setting-up-aws}
+
+Three steps. The **AWS console** path is the authoritative one; an AWS CLI equivalent is given alongside it for illustration and for administrators who automate.
+
+#### Step 1: Create the bucket {#own-lake-aws-bucket}
+
+The bucket has to live in the region the setup page names, which is the region the {{< product-c8y-iot >}} environment runs in. Provisioning compares the two and refuses a bucket anywhere else, and S3 would reject the signed request regardless: every call is signed for that one region. Cross-region access is not merely slower, it is not configured.
```
y5008720-c8y: cumbersome wording: 
- remove the unnecessary technical part ", and S3 would reject the signed request regardless: every call is signed for that one region. Cross-region access is not merely slower, it is not configured."

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+In the AWS console, go to **S3** > **Create bucket**, set **AWS Region** and set **Bucket Versioning** to **Enable**. With the AWS CLI, creating the bucket and enabling versioning are two calls:
+
+```bash
+aws s3 mb s3://<bucket> --region <region>
+
+aws s3api put-bucket-versioning --bucket <bucket> \
+  --versioning-configuration Status=Enabled
+```
+
+Use `aws s3 mb` rather than the lower-level `aws s3api create-bucket`: that one needs `--create-bucket-configuration LocationConstraint=<region>` in every region *except* `us-east-1`, where the same flag is rejected. `mb` resolves it from `--region` and is correct in all of them.
```
y5008720-c8y: remove, unnecessary AI session context (don't give an in-detail description of an approach only to say this approach should not be used)

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+```
+
+#### Step 2: Create the role {{< company-c8y >}} assumes {#own-lake-aws-role}
+
+Create the role in the **same account as the bucket**. It has two policies and both matter:
+
+|Policy|What it decides|The mistake it invites|
+|:---|:---|:---|
+|**Trust policy**|That {{< company-c8y >}}'s base principal may assume the role, and only when it presents the External ID|Naming the wrong ARN, omitting the External ID condition, or using a different External ID here than the one you enter on the setup page|
+|**Permissions policy**|What the role may do to your bucket|Granting the whole bucket where one prefix would do|
```
y5008720-c8y: Might be me, I don't understand what this table wants to describe. Instead of a column "possible mistakes" I would expect a column more like "how/what to configure".

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+**An IAM role cannot be renamed**, so getting the name wrong means deleting the role and creating it again. Create it at the default path `/` as well: a role created under a path has that path where the prefix has to be, and a role's path cannot be changed after creation either.
+{{< /c8y-admon-important >}}
+
+##### The External ID {#own-lake-aws-external-id}
+
+The External ID is **your** value, not {{< company-c8y >}}'s. The setup page pre-fills a fresh suggestion each time it is opened, so that you need not invent one, and you are free to replace it with anything you prefer. Whatever you write into the trust policy is the value that counts, and the setup page is where you state which value that was.
```
y5008720-c8y: Do I get this right that I define some "external ID" in the trust policy and need to copy that value here?
This paragraph should be rewritten, but I don't have a suggestion as I don't understand it.

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+Three things to know about it:
+
+* **Its length and character set are checked.** At least 32 characters — the suggestion is exactly that long, so keeping it is always safe. The character set is AWS's own: `A-Z a-z 0-9 _ + = , . @ : / -`, with no spaces and no `#`. A value outside it is rejected when you enter it, rather than being rejected later by AWS, where the error would surface far from its cause.
```
y5008720-c8y:
```suggestion
* **Its length and character set are checked.** At least 32 characters with a special character set `A-Z a-z 0-9 _ + = , . @ : / -`, no spaces and no `#`.
```

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+* **Use a different value for each of your {{< product-c8y-iot >}} tenants.** This is not enforced, but a value shared between two tenants removes the protection it exists to provide.
+* **Keep a record of it.** It is written into your trust policy, and the settings page never shows the value it holds — only its last few characters, enough to recognize which one is recorded.
+
+If the setup rejects an External ID, it is always one you supplied rather than the suggestion: a suggested value satisfies these rules by construction.
```
y5008720-c8y: remove

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+##### Creating the role {#own-lake-aws-create-role}
+
+In the AWS console, go to **IAM** > **Roles** > **Create role**, and choose **Custom trust policy** as the trusted entity type. Two details of that form matter:
+
+* The form opens on **AWS service**, which is the wrong choice here — that is for services inside your own account. **AWS account** is closer but still not enough: it names the account {{< company-c8y >}}'s principal lives in without letting you require the External ID. Only **Custom trust policy** lets you state both.
```
y5008720-c8y: remove this detail, duplicated information (+ change "two details matter" to "one"...)

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+To verify, read both policies back from the role:
+
+```bash
+aws iam get-role --role-name c8y-streaming-lake-ingestion-<suffix> \
+  --query 'Role.AssumeRolePolicyDocument.Statement[0].{principal:Principal,condition:Condition}'
+aws iam get-role-policy --role-name c8y-streaming-lake-ingestion-<suffix> \
+  --policy-name c8y-streaming-lake-ingestion
+```
+
+The principal must be exactly the base principal ARN from the setup page, and the External ID in the condition must be the same value you enter there — character for character. A trust policy that is otherwise correct but carries a different External ID is the most common mistake here, and it looks identical in any listing.
```
y5008720-c8y:
```suggestion
The principal must be exactly the base principal ARN from the setup page, and the External ID in the condition must be the same value you enter there. A trust policy that is otherwise correct but carries a different External ID is the most common mistake here.
```

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+##### Finding the role ARN {#own-lake-aws-role-arn}
+
+The role ARN looks like `arn:aws:iam::<your-account-id>:role/c8y-streaming-lake-ingestion-<suffix>`. In the AWS console, go to **IAM** > **Roles**, open the role and copy the **ARN** from the top of its summary panel; there is a copy button next to it, and copying the value by hand is where transcription mistakes come from. With the CLI:
```
y5008720-c8y:
```suggestion
The role ARN looks like `arn:aws:iam::<your-account-id>:role/c8y-streaming-lake-ingestion-<suffix>`. In the AWS console, go to **IAM** > **Roles**, open the role and copy the **ARN** from the top of its summary panel. With the CLI:
```

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+#### Step 3: Complete the setup in {{< product-c8y-iot >}} {#own-lake-aws-provision}
+
+<!-- SCREENSHOT: /images/datahub-guide/sli-own-lake-aws-values.png
+     Caption: "The setup page showing the base principal ARN, the suggested External ID and the required region"
+     Blocked: the setup page is not final yet. -->
+
+Open **Settings** > **Data Lake** in the **Administration** application and enter three values:
+
+* **Role ARN** — as retrieved above.
+* **External ID** — the value you actually put in the trust policy, whether that is the suggestion or your own. Read it back from the role rather than from your notes: **IAM** > **Roles** > your role > **Trust relationships**.
```
y5008720-c8y:
```suggestion
* **External ID** — the value you put in the trust policy. To be found here: **IAM** > **Roles** > your role > **Trust relationships**.
```

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+* **Bucket** and **prefix** — which together form the base location `s3://<bucket>/<prefix>`. See [Choosing a base location](#own-lake-base-location).
+
+The region is not one of them, because the setup page gave it to you in the first place. Do confirm that the bucket landed there, since a mismatch fails the setup.
```
y5008720-c8y: remove, duplicate info

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+<!-- SCREENSHOT: /images/datahub-guide/sli-own-lake-aws-input.png
+     Caption: "Entering the role ARN, External ID, bucket and prefix"
+     Blocked: the setup page is not final yet. -->
+
+Then start the setup. It does not start on its own: nothing watches your AWS account for the role, because only you know when it is actually in place. What happens next is described in [What the setup verifies](#own-lake-verification).
```
y5008720-c8y:
```suggestion
Then start the setup. What happens next is described in [What the setup verifies](#own-lake-verification).
```

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+### Setting up your Azure storage account {#own-lake-setting-up-azure}
+
+Four steps. The **Azure portal** path is the authoritative one; an Azure CLI equivalent is given alongside it.
+
+Everything happens in the **storage Entra directory** — the directory that your storage account's subscription belongs to. Azure only allows a storage account to be granted to a service principal that exists in that storage account's own directory. An approval recorded in any other directory cannot even be found in the storage account's access-control picker, and the directory ID you enter in step 4 has to be that same one.
```
y5008720-c8y:
```suggestion
Everything happens in the **storage Entra directory** — the directory that your storage account's subscription belongs to. Azure only allows a storage account to be granted to a service principal that exists in that storage account's own directory.
```

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+#### Step 1: Approve the Streaming Lake Ingestion application {#own-lake-azure-consent}
+
+You are not creating an application here: {{< company-c8y >}}'s already exists, in {{< company-c8y >}}'s own directory. What you are doing is **recording that your directory trusts it**, which Azure represents as an enterprise application in your directory. It grants access to nothing. It only creates the object that step 2 assigns roles to, and until then the application can do nothing in your directory at all.
```
y5008720-c8y: not sure this paragraph is necessary

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+{{< c8y-admon-info >}}
+**The application requests no API permissions.** It cannot read your directory, its users, or any other resource. The only thing it can ever do is what you grant it with Azure RBAC in step 2.
+{{< /c8y-admon-info >}}
+
+In the Azure portal, open the consent link shown on the setup page, signed in as an administrator **of the storage Entra directory**. Review the prompt: it names the application and states that no permissions are requested. **Check the directory name shown on the sign-in and consent screens** — if you hold accounts in several directories, the browser may already be signed in to a different one, and approving there records the approval where it cannot be used. Then accept.
```
y5008720-c8y:
```suggestion
In the Azure portal, open the consent link shown on the setup page, signed in as an administrator **of the storage Entra directory**. Review the prompt: it names the application and states that no permissions are requested. Make sure you are signed in to the correct directory, then accept.
```

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+The Azure CLI path produces the same object with no browser and no consent prompt, which helps when the consent prompt is blocked by policy or when your organization automates directory changes:
+
+```bash
+az login --tenant <storage-entra-directory-id>
+az ad sp create --id <client-id>
+az ad sp show --id <client-id> --query "{name:displayName, appId:appId, objectId:id}" -o table
+```
+
+The last call must return one row whose `appId` equals the client ID. Confirm which directory the commands ran against with `az account show --query tenantId -o tsv`; a portal session tells you nothing about where the CLI is pointed.
```
y5008720-c8y:
```suggestion
The last call must return one row whose `appId` equals the client ID. Confirm which directory the commands ran against with `az account show --query tenantId -o tsv`.
```

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+```
+
+The last call must report exactly these two. The Data Contributor's scope must end in `/containers/<container>`, the Delegator's in `/storageAccounts/<account>`.
+
+Role assignments are not effective instantly. If step 4 fails immediately after this, wait a few minutes and retry before changing anything.
+
+#### Step 3: Enable soft delete {#own-lake-azure-soft-delete}
+
+{{< c8y-admon-req >}}
+Enable **blob soft delete** and **container soft delete** on the storage account before you use Streaming Lake Ingestion in production. Each keeps deleted data recoverable for a retention window you choose. Without them a delete is immediate and final: an accidentally deleted container, or a delete run against the wrong path, takes the tables and their metadata with it, and nothing can bring them back. The service level agreement requires this for production use, so treat it as a step to complete rather than an option to weigh.
```
y5008720-c8y:
```suggestion
Enable **blob soft delete** and **container soft delete** on the storage account before you use Streaming Lake Ingestion in production. Each keeps deleted data recoverable for a retention window you choose. Without them a delete is immediate and final: an accidentally deleted container, or a delete run against the wrong path, permanently deletes tables and their metadata. The service level agreement requires this for production use.
```

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+    --enable-delete-retention true --delete-retention-days 7 \
+    --enable-container-delete-retention true --container-delete-retention-days 7
+
+az storage account blob-service-properties show --account-name <account> \
+    --query "{blobs:deleteRetentionPolicy, containers:containerDeleteRetentionPolicy}"
+```
+
+Both policies must report `enabled: true` with the retention you set. Seven days above is an example; choose a window that matches how long you would want to be able to recover.
+
+Provisioning does not check this setting, so an account without soft delete provisions and runs normally. Confirming that it is on is part of this step, not something you will be warned about later.
```
y5008720-c8y: remove

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+  # the directory that subscription belongs to — this is the value to enter
+  az account show --subscription <sub> --query tenantId -o tsv
+  ```
+
+  `az account show --query tenantId` on its own reports the directory of whichever subscription is currently selected, which stops being the right answer the moment you have access to more than one.
+
+* **Base location** — `abfss://<container>@<account>.dfs.core.windows.net/<path>`. See [Choosing a base location](#own-lake-base-location).
+
+Then start the setup. It does not start on its own: nothing watches for the role assignments from step 2, because only you know when they are actually in place.
```
y5008720-c8y:
```suggestion
Then start the setup.
```

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+### Choosing a base location {#own-lake-base-location}
+
+The base location is the path your tenant's data is written under. {{< product-c8y-iot >}} appends the tenant's ID to it, so the tenant's tables live under `<base location>/<tenant-id>/`.
+
+Three rules apply on both clouds:
+
+* **Several tenants may share one prefix.** Because each tenant's data lands one level down, under its own ID, two tenants given the same base location end up in sibling directories rather than on top of each other. Pointing your tenants at one common prefix is a supported layout.
+* **What is refused is one tenant nested inside another's.** If the location your tenant would write to contains, or is contained by, one another tenant on the environment already holds, the setup refuses it and asks you to pick another — nothing is created, and your storage is not touched. For privacy the refusal does not name the other tenant, since it may belong to a different customer. Choose a location that neither contains nor sits inside the other one.
```
y5008720-c8y:
```suggestion
* **What is refused is one tenant nested inside another's.** If the location your tenant would write to contains, or is contained by, one another tenant on the environment already holds, the setup refuses it and asks you to pick another — nothing is created, and your storage is not touched. Choose a location that neither contains nor sits inside the other one.
```

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+* **The base location is fixed once the setup has succeeded.** It is recorded when your tenant's Iceberg catalog is created, and running the setup again does not rewrite it. Entering a different one later therefore moves nothing: the tables stay where they are, and anything already written under the old path would be stranded rather than migrated. [Contact {{< company-c8y >}} support](/additional-resources/contacting-support/) to change it. The same holds for the storage type and, on Azure, the Entra directory.
```
y5008720-c8y:
```suggestion
* **The base location is fixed once the setup has succeeded.** It is recorded when your tenant's Iceberg catalog is created, and can not be changed afterwards. [Contact {{< company-c8y >}} support](/additional-resources/contacting-support/) if you need to change it. The same holds for the storage type and, on Azure, the Entra directory.
```

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+* **Saving re-runs the verification.** The new role is assumed, and a write, a read-back and a delete are exercised under your base location, so a trust policy that no longer admits {{< company-c8y >}} is reported there and then rather than breaking ingestion later.
+* **Rotate the External ID in your trust policy first, then save it here.** The two have to agree: while they do not, every assume-role call fails and Iceberg commits stop. Leaving the field blank keeps the stored value, so you can replace the role alone.
+* **The replacement role has to be in the same AWS account.** That account is fixed when your catalog is created, because your bucket lives in it, and a role in another account is refused with a message naming both. Moving the data lake to a different account is a change to raise with support.
+
+### What the setup verifies {#own-lake-verification}
+
+The setup does not take your word for the grant. It exercises it, end to end, through the identity that will actually be used — rather than inspecting your cloud account's settings, which this onboarding gives it no permission to read:
```
y5008720-c8y: change wording

content/datahub/streaming-lake-ingestion-bundle/own-data-lake.md
```diff
@@ -0,0 +1,528 @@
...
+
+1. **The grant.** On AWS it assumes your role with your External ID; on Azure it acquires a token for your directory as the consented application.
+2. **The data path.** With those credentials it lists, writes, reads back and deletes a test object under the base location.
+3. **The catalog.** Your tenant's Iceberg catalog is created against the values you entered.
+4. **Credential vending.** It then checks that the catalog can hand out the short-lived credentials queries need — a scoped session on AWS, a user-delegation token on Azure. This step comes last because credentials are vended per table, so there has to be a table to ask about: the check creates an empty one called `vending_check`, in a namespace called `iceflow_provisioning_check`, and requests credentials for it.
+
+    That table is created once and **left in place**, rather than dropped after each run: dropping it without purging would abandon its metadata in your storage every time, where nothing would later clean it up. It stays empty and costs nothing. If you find it while auditing your storage, that is what it is — leave it, and later runs reuse it.
```
y5008720-c8y: remove

y5008720-c8y (changes requested):
find some comments below

PR #5154, opened 2026-09-14
[DM-7271] Add Deployment Manager user documentation

dosc-c8y:
Adds the user documentation for Deployment Manager, the Device Management feature that rolls firmware, software and configuration out to large fleets of devices.

The Deployment Manager UI plugin contributes a **Deployments** entry to the Device Management navigator, and its list view carries a contextual help panel. That panel holds no text of its own: `c8y-help` fetches the `helpcontent` entry named by its `src` from this page, so the panel and its **Open documentation** button lead to the same words. The plugin points at `/docs/device-management-application/deployments/#deployments`.

The page body is the documentation behind that button.

### Please look at this first

The **Smart suggestions** bullet comes from the UI design and describes a feature that has **no backend yet** — nothing records the deployment keys that devices asked for and did not find. Either that ships before this page is published, or the bullet is cut from both the `helpcontent` block and the body. I kept it because the design is the agreed source for the panel text, but it should not go live as written.

### Notes for review

- The help text is a YAML block scalar (`content: |`), as in `enterprise-tenant/customization.md`, rather than the double-quoted form most `helpcontent` blocks use. In the quoted form a single newline folds to a space, which merged a subheading into the following paragraph; the block scalar renders the paragraphs as intended. Verified by building and diffing the generated `index.json`.
- Built locally with `hugo` — 9548 pages, no errors — and the generated `index.json` checked for the panel content.
- Jira: DM-7271, the story that adds the deployment list to the UI.

AI usage: written and reviewed with Claude Code (Opus 5).


dosc-c8y:
Closed for now. While deployment-manager is under active development and awaiting release, we will provide UI context help through an alternative method. This PR is only noise currently.

PR #5155, opened 2026-09-14
Fix links

BeateRixen:
Fixed various broken links found by the link checker. One issue remains to be resolved left (see comment).

PR #5165, opened 2026-09-16
Fix broken links in develop (alternative to #5155)

pawel-rynarzewski-c8y:
Fixes the broken links reported for `develop` by Link Checker run [34812160055](https://github.com/Cumulocity-IoT/c8y-docs/actions/runs/34812160055/job/103875509266) (2026-09-14).

This is a **self-contained alternative to #5155** for the same set of failures, so the two approaches can be compared before deciding which to merge. It does not depend on #5155 being merged. `develop` only — backporting to be decided separately.

## What was failing

| # | URL | Root cause | Fix |
|---|-----|-----------|-----|
| 1 | `codex/components/application-and-system/internationalization/overview#annotating-texts…` | Codex page moved to `common-tasks/`; old path renders "404 Not found" | Repointed |
| 2 | `ecosystem.cumulocity.com/cumulocity-iot-ecosystem/devices/` | Time-windowed block from the CI runner | **Deliberately not changed** — see below |
| 3 | `change-logs/#ui-c8y-**DM**-7021-deprecate-availability-donut-charts` | Anchor case mismatch | Rename + link fix |
| 4 | `javadoc.io/static/io.rest-assured/rest-assured/**3.0.0**/…#content-byte:A-` | javadoc.io no longer serves 3.0.0 at all | Repointed to 3.3.0 |
| 5 | `change-logs/…#apama-in-c8y-20260126-**Add**-logger-block-in-public-preview` | Same anchor case mismatch | Rename + link fix |

#5 was **not** in the failing run — it was merged in `f59b2c942` a few hours afterwards and would have failed the next weekly run. Fixed pre-emptively.

## Root cause of #3 and #5

`themes/c8ydocs/layouts/partials/change-logs-list.html:158` builds each entry's anchor with:

```go
id='{{(replace .Name ".md" "") | urlize }}'
```

`urlize` **lowercases**, so the rendered id is the lowercased source filename. Authors naturally copy the filename's own casing (a ticket ID like `DM-7021`) into the link, and since browsers match `id` fragments case-sensitively, the link silently lands at the top of the page rather than the entry.

Confirmed against the live page: **no section id on `/docs/change-logs/` contains a single uppercase character.**

```
$ curl -s https://cumulocity.com/docs/change-logs/ | grep -oE "id='[a-z0-9-]*[A-Z][^']*'"
(no output)
```

See for yourself — these differ only in `DM` vs `dm`:

- broken: https://cumulocity.com/docs/change-logs/#ui-c8y-DM-7021-deprecate-availability-donut-charts (stays at top)
- works: https://cumulocity.com/docs/change-logs/#ui-c8y-dm-7021-deprecate-availability-donut-charts (jumps to *Deprecation of device availability views*)

**The renames here are anchor-neutral.** `urlize` had already lowercased the id, so the published anchor is unchanged and no existing link breaks — the rename only stops the filename from misleading the next author.

> This is where this PR differs from #5155, which instead renamed the file to drop `DM-7021` entirely. That changes an already-published public anchor, and cannot pass the link checker until the site is redeployed — which is why the link is still red on that PR. The checker validates against live production, so an anchor for a change-log entry added in the same PR can never pass.

The other 297 change-log files with uppercase basenames are left alone; they are harmless until someone links to them. Worth a separate discussion.

## #2: why `ecosystem.cumulocity.com` is deliberately not excluded

`https://ecosystem.cumulocity.com/cumulocity-iot-ecosystem/devices/` failed all 10 retries at exactly the 10s `requestTimeout` (06:33–06:35), while a targeted CI diagnostics run 3h later returned 200 in 659 ms and curl returns 200 in 0.8 s. A time-windowed block, not a broken link.

We deliberately do **not** add it to `excludedLinks`: it is our own domain, and excluding our own domains hides exactly the links we most need checked. If this keeps flaking, the fix belongs in our infrastructure — letting the link checker reach our own sites reliably from CI — not in the checker's exclusion list.

This link may therefore still flake. That is the accepted outcome here, not a regression.

## Checker improvement

A case-mismatched fragment previously reported only *"An element with id … should exist"*, giving no hint that the fix was a single character — which is what sent #5155 down the renaming path. The checker now names the near-match:

```
Error: An element with id or name = "ui-c8y-DM-7021-deprecate-availability-donut-charts"
should exist in HTML or frames - but "ui-c8y-dm-7021-deprecate-availability-donut-charts"
exists, which differs only by case, so the link's fragment casing is wrong
```

Generic to every site the checker visits, not just change-logs.

## Verification

Each fixed link re-checked locally with `node run.cjs --urls-with=…` — all pass. The Codex fix was confirmed in a real browser (it is a client-rendered SPA, so `curl` returns an identical shell for both paths): the old path's title outlet reads "404 Not found", the new one reads "Internationalization" with the fragment present.

The `broken-link-check` label is applied to this PR for a full CI run.


content/microservice-sdk/java-bundle/developing-microservice.md
```diff
@@ -986,7 +986,7 @@ The following locations are searched for the logback configuration file:
 
 A Spring Boot library was upgraded to 2.5.8, hence upgrading Microservice SDK to 10.13+ may require some additional development.
 
-* The `content(matcher)` method of RestAssured has been replaced with `body(matcher)`, see [RequestSpecification#content()](https://javadoc.io/static/io.rest-assured/rest-assured/3.0.0/io/restassured/specification/RequestSpecification.html#content-byte:A-)
+* The `content(matcher)` method of RestAssured has been replaced with `body(matcher)`, see [RequestSpecification#content()](https://javadoc.io/static/io.rest-assured/rest-assured/3.3.0/io/restassured/specification/RequestSpecification.html#content-byte:A-)
```
pawel-rynarzewski-c8y, line 989: @mgrumann javadoc.io no longer serves rest-assured 3.0.0 at all — `/static/.../3.0.0/...` returns 404 with an empty body, and `/doc/.../3.0.0/...` redirects to the version list — so this link was dead.

I repointed it to **3.3.0**, the nearest version javadoc.io still hosts that documents `content(byte[])` (4.x and 5.x dropped the method, so their pages have no `#content-byte:A-` anchor). Kept the `/static/` form deliberately: it serves the javadoc directly and the browser jumps to the method, whereas the `/doc/` wrapper loads it in an iframe without forwarding the hash.

Could you confirm 3.3.0 is an acceptable reference for this Microservice SDK 10.13+ upgrade note, or whether it should point elsewhere? The original author (Kamil Sobala, MTM-42401, 2022) is no longer with us.

cc @BeateRixen

BeateRixen:
@pawel-rynarzewski-c8y I actually already created a PR with the other link fixes (except for the one with the timeout) :-). https://github.com/Cumulocity-IoT/c8y-docs/pull/5155. I will close my PR to avoid confusion.

mgrumann:
Here the Okay form my side

mgrumann (approved):

BeateRixen (approved):

pawel-rynarzewski-c8y:
> @pawel-rynarzewski-c8y I actually already created a PR with the other link fixes (except for the one with the timeout) :-). #5155. I will close my PR to avoid confusion.

@BeateRixen, I was aware of your PR, what I proposed here is an alternative approach that preserves the existing changelog file names. The actual problem was the letter case in the links, not the fact that the changelog names contain ticket numbers. On top of that, this PR adds a clearer error message for the future, for cases where the correct anchor exists but differs only in letter case.

BeateRixen:
> > @pawel-rynarzewski-c8y I actually already created a PR with the other link fixes (except for the one with the timeout) :-). #5155. I will close my PR to avoid confusion.
> 
> @BeateRixen, I was aware of your PR, what I proposed here is an alternative approach that preserves the existing changelog file names. The actual problem was the letter case in the links, not the fact that the changelog names contain ticket numbers. On top of that, this PR adds a clearer error message for the future, for cases where the correct anchor exists but differs only in letter case.

@pawel-rynarzewski-c8y I was aware that the capital letter was the problem. I just removed the ticket number entirely as we never use it in changelog file names. But you are right, for future cases, this is the clearer approach.

PR #5178, opened 2026-09-21
Fix sector-level PDF (ZIP) download being deleted on every deploy

pawel-rynarzewski-c8y:
## Problem

The sector-level "Download PDFs" button (a ZIP of every PDF in a section) intermittently 404s. Reported as develop-only, but it affects the release branches too — develop just hits it far more often because it deploys on every merge.

In `staging.yml`, the first upload step excluded `pdfs` but not `zips`:

```
rsync ... --exclude='pdfs' ./public/ <host>:<dest> --delete-after
```

`public/zips/` is produced *later* in the job by the PDF generator, so it does not exist yet at that point and `--delete-after` deletes the published sector ZIPs from the server. They only come back ~17 minutes later in the final step. A run cancelled in between leaves them gone until the next green deploy.

Confirmed in [run `35218893305`](https://github.com/Cumulocity-IoT/c8y-docs/actions/runs/35218893305/job/105199676456) (`Upload documentation` and `Upload ZIPs` steps of the prod job):

```
12:36:56  Upload documentation (without PDFs)  deleting zips/
12:54:23  Upload ZIPs                          created directory .../cd/zips
```

And observed live on 2026-09-21 while deploys were in flight — every sector ZIP 404, every individual PDF 200, on both `develop` and `release/y2026`.

## Changes

**`.github/workflows/staging.yml`**

- `--exclude='zips'` alongside `--exclude='pdfs'`, so a deploy no longer takes the published ZIPs offline. This is the actual fix.
- The four PDF steps gated on `if: matrix.env == 'prod'`. The `staging-resources` host is behind HTTP basic auth, so wkhtmltopdf fails every render there with `Authentication Required`, uploads empty PDFs and empty ZIPs, and burns ~15 minutes (see https://github.com/Cumulocity-IoT/c8y-docs/actions/runs/35585682077/job/106288353553 - Generate PDFs step: 178 `Authentication Required`, 29 failed PDFs). It has never produced a usable PDF, and it is not develop-specific - the same run on [release/y2026](https://github.com/Cumulocity-IoT/c8y-docs/actions/runs/35218904629/job/105198920668) and an earlier one on [develop](https://github.com/Cumulocity-IoT/c8y-docs/actions/runs/35218893305/job/105199676621) both show 167 auth errors and 26 failed PDFs, while the `prod` job of the same run ([example](https://github.com/Cumulocity-IoT/c8y-docs/actions/runs/35585682077/job/106288353843)) reports none. The step then uploads the result anyway: `sent 67 bytes` for `pdfs/` and `sent 643 bytes` for six empty ZIPs, with `--delete-after`.
- `--delete-after` on the PDF/ZIP uploads is now conditional on the generator reporting a complete run. Otherwise a single failed render would delete the live file and recreate the same 404.

**`pdf-generator/pdf-generator.ts`** — it exits 0 on failures so a
transient render error does not block a docs release, which is unchanged: the old code already
caught the `execSync` error and discarded the result, so a failed render never failed the
workflow either. The three changes below exist to protect us from overwriting existing files with incomplete ones.

- A sector with **no** available PDFs gets **no ZIP written**, instead of an empty archive being
  published over the working one. Partial sectors still get a ZIP with whatever exists.
- The generator reports `complete` to the workflow, which is what gates the conditional
  `--delete-after` above.
- A bundle whose content pages **all** failed to load counts as failed. This one is not
  cosmetic: wkhtmltopdf writes a cover/copyright/TOC-only PDF in that case and does **not**
  reliably exit non-zero (verified — an unreachable content URL yields a 22 KB, 4-page file and
  exit code 0). Without the check, that empty shell would be published and the run would report
  `complete=true`, authorising `--delete-after` to remove the real PDFs.

## Verification

`tsc --noEmit` passes. Run locally against a production Hugo build served over HTTP, exit 0 throughout:

| Scenario | ZIP | `complete` |
| --- | --- | --- |
| Content URL unreachable (wkhtmltopdf exits 0, cover-only PDF) | not written | `false` |
| Good bundle + all-404 bundle | written, 1/2 PDFs | `false` |
| Everything renders | written, 1/1 PDFs | `true` |

To confirm after merge:

- `Upload documentation` no longer logs `deleting zips/`.
- `curl -sI https://cumulocity.com/docs/zips/device_management.zip` returns 200 *while a build is still running* — the case that is broken today.
- `staging-resources` shows the four PDF steps skipped and finishes ~15 min faster; `prod` still runs all four and reports `complete=true`.

## Notes

- With conditional pruning, a genuinely removed bundle's PDF lingers on the server until the next fully green deploy. Nothing links to it, so it is invisible — a deliberate trade against the site breaking.
- **Both release branches have the same defect** and are worth a backport. `release/y2025` carries an identical `staging.yml` (only `pdfs` excluded) and the same ZIP feature, so it is affected too - its ZIPs happen to be up right now only because its last deploy (2026-09-17) finished cleanly. It deploys rarely, so the 404 window is almost never open there; `release/y2026` deploys often enough to hit it.


BeateRixen (approved):

reey (approved):

pawel-rynarzewski-c8y:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|release/y2026|[](https://github.com/Cumulocity-IoT/c8y-docs/pull/5186)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport)

pawel-rynarzewski-c8y:
## 💚 All backports created successfully

| Status | Branch | Result |
|:------:|:------:|:------|
|✅|release/y2025|[](https://github.com/Cumulocity-IoT/c8y-docs/pull/5187)|

### Questions ?
Please refer to the [Backport tool documentation](https://github.com/sorenlouv/backport)

